PR #112, opened 2016-11-22
Insert CSS & JS to avoid document.write blocking

bshouse:
The CSS and JS injections happening with document.write gets blocked due to cross-domain loading. Used an alternate method to allow loading.


bshouse:
Here is some notes on the problem/solution- http://stackoverflow.com/questions/39610829/a-parser-blocking-cross-origin-script-is-invoked-via-document-write-how-to-ci#

bshouse:
Closing to work on web branch instead

PR #113, opened 2016-11-22
Migrate to official npm package name

bshouse:
electron-prebuilt is a legacy package name that made it difficult for developer to find Electron with npm. electron-prebuild still exists for backward compatibility, but developers are encouraged to migrate. https://github.com/electron-userland/electron-prebuilt/issues/160


bshouse:
Closing to work on the web branch instead.

PR #144, opened 2016-12-27
merge photo upload and dropbox support

larrysalibra:


shea256:
Great great work on this Larry. I'll follow up more on the journal.

I only had one issue... for some reason "fetch" (AKA "proxyFetch") isn't working for me with the name registration operation (I receive a CORS error). However, if I switch it to "realFetch" (AKA "fetch") then suddenly the registration goes through. Any idea what's happening here?

screen shot 2016-12-27 at 5 55 19 pm

larrysalibra:
Thanks @shea256 - opened an issue for this #146

PR #360, opened 2017-03-29
v0.6 core auth

jcnelson:
This PR adds support for generating Core session tokens once the application authentication request has been approved.  The core session token gets sent to the app as part of an auth response.

This requires the `develop-core-auth` branch of `blockstack.js` to work.


larrysalibra:
I'll need to update AuthModal to only try to get a core session token if the auth response has a `domain_name`

PR #442, opened 2017-04-26
wallet "deposit" url was wrong

dominictarr:
should be `/wallet/receive`


dominictarr:
oh sorry, this should be againts 0.7

PR #444, opened 2017-04-28
Update README.md for accuracy and clarity

jeffslofish:
Simple fixes to the documentation. Made the subheadings for Linux the same heading level. Fixed all the places where items weren't being displayed because they were treated as HTML entities.


larrysalibra:
Thank you for the PR @jeffslofish ! I've merged it in to `v0.8`

Issue #320, opened 2017-03-17
resolve flexbox flow orientation on home screen


guylepage3:
There seems to be an issue with the flow of the flexbox orientation on the home screen when an even number of app icons appear.

guylepage3:
<img width="838" alt="screenshot 2017-03-17 10 36 17" src="https://cloud.githubusercontent.com/assets/1711854/24047999/dbe5d874-0afd-11e7-9f11-a5c19dec4059.png">

matheusml:
By replacing ```justify-content: space-between``` to ```justify-content: flex-start```, we get this result:

![](http://i.imgur.com/iNJgfUd.png)

Can I send the PR?

guylepage3:
Awesome @matheusmi! Feel free to do a pull request.

guylepage3:
resolved. closing issue.

PR #590, opened 2017-07-04
Fixing flexbox orientation

matheusml:
#320


guylepage3:
@matheusml can you please pull from `ui-0.11` branch and I'll merge it in?

matheusml:
Done!

larrysalibra:
Thanks @matheusml !
@guylepage3 you take a look at this and merge it into your design branch if appropriate?

guylepage3:
@larrysalibra @matheusml taking a look at this now

guylepage3:
Ok. so we're actually looking to execute something a bit different here @matheusml. If you see below, you've adjusted `flex-box`'s justification to be `flex-start` which does align objects left to right and flow down. but what we are looking to execute is to center objects and have them flow outward.

Would you like to take another stab at it? 😄 

screenshot 2017-07-07 11 35 17

Issue #607, opened 2017-07-11
Setting owner key from browser results in registrations with wrong address


kantai:
It looks like the issue I referenced at the end of blockstack-core issue #487 (the address being registered to by core is wrong), is that browser is expecting a compressed bitcoin address, while core is registering with an uncompressed address.

This is because the private key string passed to core needs `01` suffix indicating that it should get a compressed public key (see WIF format here: [https://en.bitcoin.it/wiki/Wallet_import_format]).

Anyways, I have a patch which fixes this issue, and has been tested in regtest. Creating PR now.

larrysalibra:
Fixed in `v.0.11` Thank you @kantai for figuring this out!

PR #608, opened 2017-07-11
adding compression suffix to the 'owner set key' operation

kantai:
Issue #607


larrysalibra:
Thanks for this @kantai - I tested this out as well on regtest and it works well for me.

I updated tests for this pull request - in the future it'd be great if pull requests passed the CircleCI tests. :-)

Very excited to get this plus `unsafe` into the hands of users!

Issue #616, opened 2017-07-13
flex-box issue in safari


guylepage3:
@brendanmmcmanus sent in a screen shot this morning showing an issue with the alignment of the app icons on the home screen of the browser and the following trouble shooting conversation occurred.

Brendan McManus [8:44 AM] 
![pasted image at 2017_07_13 08_44 am](https://user-images.githubusercontent.com/1711854/28167201-c811e1e6-67a8-11e7-8124-4746d3111978.png)

Brendan McManus [8:44 AM] 
when I change the browser window size, they tile normally

Brendan McManus [8:44 AM] 
![pasted image at 2017_07_13 08_44 am 1](https://user-images.githubusercontent.com/1711854/28167216-d57af732-67a8-11e7-83bf-ba4e71975790.png)

Brendan McManus [8:44 AM] 
but at a certain width they don’t

Guy Lepage [8:44 AM] 
interesting.. I am unable to replicate this on my end
are you on a retina display?
also, can i ask you to inspect element?

Brendan McManus [8:45 AM] 
I’m using safari version 11 on high sierra beta, and yeah retina with a monitor
yup one sec

Guy Lepage [8:46 AM] 
on this div.. `div class="container app-wrap"`
ok let me view this in safari
could be a `flex-box` issue with safari

Brendan McManus [8:47 AM] 
![pasted image at 2017_07_13 08_47 am](https://user-images.githubusercontent.com/1711854/28167261-0186065a-67a9-11e7-8b5d-dc5d309ae72c.png)

Guy Lepage [8:47 AM] 
interesting.. it's an issue with safari..
i have been able to replicate the issue.

matheusml:
should we leave flexbox and try to use bootstrap grid system, for example?

guylepage3:
@matheusml the one thing i do like about having flex-box in now is that if someone coming in from the core team needs to add something it's already there so they will have something to roll with.

guylepage3:
complete

PR #626, opened 2017-07-17
Fixed Flexbox for Safari

estermer:
Closes: #616 

There was an issue with `display: flex` unnecessarily being declared within 3 nested elements when it was only needed in the inner most element. Tested on both Chrome and Safari at all viewport sizes including mobile and it is now working smoothly.

as per the discussion in the ticket, I was able to maintain using flexbox rather than something else.

screen shot 2017-07-17 at 3 21 22 pm


guylepage3:
@estermer can you quickly resolve this pull request and let me know when it's complete by @ messaging me in slack? Thank you sir. :)

estermer:
@guylepage3 all set

larrysalibra:
Thanks @estermer for this and @guylepage3 for reviewing!

Issue #638, opened 2017-07-19
spacing issue in wallet


guylepage3:
Address spacing issue in the wallet

<img width="1193" alt="screen shot 2017-07-19 at 10 24 29 am" src="https://user-images.githubusercontent.com/1711854/28383968-f328f3c8-6c90-11e7-8738-89e54c2d4dfc.png">

jackzampolin:
This has been fixed in later releases. Closing.

Issue #566, opened 2017-06-27
profile list card pending is smaller than non-pending card


guylepage3:
Profile list card pending is smaller than non-pending card

<img width="1095" alt="screenshot 2017-06-27 15 11 20" src="https://user-images.githubusercontent.com/1711854/27605527-fd5d771c-5b4a-11e7-8859-9a8c7ea3c0b2.png">

guylepage3:
resolved.

Issue #395, opened 2017-04-05
Profile pictures shape


larrysalibra:
If you don't upload a square image, avatars aren't circles:

<img width="751" alt="screen shot 2017-04-06 at 2 37 51 am" src="https://cloud.githubusercontent.com/assets/597182/24721210/29053108-1a72-11e7-8a0a-5ef64cb4fb1d.png">

jackzampolin:
This has been fixed in the latest version

PR #639, opened 2017-07-20
Interface bug fixes for wallet and profile

yknl:
**Fix profile pictures shape. #395** 
Fixed using `object-fit: cover`. Not supported on IE though, image will be stretched to fit instead.

**Fix profile list pending card smaller than non-pending card. #566** 
profile-list-card-size

**Fix wallet app extra white space at bottom of page issue. #638** 
wallet-space-issue


larrysalibra:
Thank you @yknl

PR #667, opened 2017-08-03
Update manual-core-to-browser-migration.md

oluapocram:
Updated documentation to disable development mode after copying Core API Password.


larrysalibra:
Thank you!

Issue #678, opened 2017-08-12
More accurately specify node version


larrysalibra:
I'm using node 6 LTS - we've specified node >= to 6 in the package.json file. For whatever, people have reported that node 8 doesn't work. We should update package.json to specify the range of node versions that are known to work or alternatively fix whatever isn't working on node 8.

PR #679, opened 2017-08-13
specify range of node versions

glambeth:
Update engines field in package.json per #678


larrysalibra:
Thank you @glambeth !

Issue #644, opened 2017-07-23
Add default profile feature


larrysalibra:
Provide a mechanism for users to use a different profile for sign in with blockstack when they are prompted to approve a sign in request and from the list of profiles screen.

---
Users currently can have multiple profiles, but we have hardcoded authentication to use the first profile. As such, there's no way to sign in to a blockstack app with anything but the first profile.

<img width="1180" alt="screen shot 2017-08-11 at 12 14 42 am" src="https://user-images.githubusercontent.com/597182/29180411-54f7a87e-7e2a-11e7-8983-3337826d3920.png">

We'd like to add a mechanism to set a profile as the default profile on the above screen. The default profile is the one that will be used for authentication unless they override the choice.

We'd also like to give users the ability to choose another profile at sign in approval in the screen below and indicate what profile they're using to sign into the app:

<img width="433" alt="screen shot 2017-08-11 at 12 15 42 am" src="https://user-images.githubusercontent.com/597182/29180717-374e3e9a-7e2b-11e7-986b-7992b178d383.png">

amaurymartiny:
3 questions:

- There's already a `current` field in `profiles.identity` redux state, which I assume denotes the current active profile. It has duplicate data as one of the `localIdentities`, so would it be a good idea to replace this `current` object into a `currentDomainName` string? That string would map to the corresponding `localIdentities` item. (No big deal, we leave that maybe for another PR).
-  I thought this `current` field was the one used in the Sign In Request modal, but actually as you said Larry it's hardcoded to use the 1st profile. So my question is: where is the `current` object used?
- Would we want to add another `default` field in the state, or do I update the `current` field when user chooses his default profile (issue title is pretty confusing on this)?

---

On the 3rd question: UX-wise, the main difference I see is the following. On the "Sign In Request" modal (screenshot above):

- If we have both `default` and `current` fields, on this modal, it would always be the default profile that shows up first (can be overridden). 
- If we only have a `current` field like now, every time this modal shows up, current profile shows up. If user chooses another profile in this modal, the current profile will be updated everywhere (including in the user settings). Should probably change wording Default->Current in this case.

larrysalibra:
@amaurymartiny I believe the the `current` object is used to display the `currently` viewed profile `ViewProfilePage`. I agree this is confusing - I changed to issue title to be "default". 

[...]

PR #680, opened 2017-08-13
Add Default Profile

amaurymartiny:
Implements #644.

2 things:
- The UI is pretty basic now. I added a `<select>` dropdown to choose your profile in the AuthModal (default one is of course selected by default). UX can be discussed (a [Collapse](http://v4-alpha.getbootstrap.com/components/collapse/#example) to select a different profile? Another page to select your profile, like Google?)
- I added a piece of code for retro-compatibility, i.e. if the user has `localIdentities` but not a default one, then the 1st one is selected by default. Or do we want to add some UX to prompt the user to choose his default one?


larrysalibra:
> I added a piece of code for retro-compatibility, i.e. if the user has localIdentities but not a default one, then the 1st one is selected by default. Or do we want to add some UX to prompt the user to choose his default one?

I think this is fine.

> The UI is pretty basic now. I added a <select> dropdown to choose your profile in the AuthModal (default one is of course selected by default). UX can be discussed (a Collapse to select a different profile? Another page to select your profile, like Google?)

For the purposes of the issue, I think this is good. We can rework the design.


@amaurymartiny This works well when users don't have a name associated with their profile, but it breaks after adding a username.

This line in particular:
https://github.com/blockstack/blockstack-browser/pull/680/files#diff-44d6650504a521b77e2c6397ab1914f5R113

After adding a username, userDomainName is the username instead of the address.

Not sure if you saw this, but there's a way to set up a blockstack core node in regtest mode which will give you ~50 regtest "bitcoins" and let you register names very quickly for testing.

There's a document on how to do that here -> https://github.com/blockstack/blockstack-core/blob/master/docs/interactive_regtest_macos.md

Instead of using the virtualenv from the branch it specifies in the doc, I'd suggest using one of the CircleCI artifacts from the master branch, such as this one: https://387-24375248-gh.circle-artifacts.com/0/tmp/circle-artifacts.oDp01rU/blockstack-venv.tar.gz

What you'll want to do on the mac app is:

1) enable development mode
2) enable regtest mode (this shuts off the bundled core node)
3) start the regtest node from the command line

amaurymartiny:
Oh ok I didn't know about this, thanks!

Will fix this asap.

larrysalibra:
Thank you very much @amaurymartiny! It works perfectly for me now!

Issue #579, opened 2017-06-30
ability to open blockstack automatically when you log in (macOS Login Items)


guylepage3:
It would be extremely helpful to add a Blockstack Preferences Pane that would automatically add "able to open automatically when you log in" located in macOS `System Preferences / Users & Groups/ User / Login Items`.

<img width="299" alt="screenshot 2017-06-30 08 33 55" src="https://user-images.githubusercontent.com/1711854/27735798-554ead8a-5d6f-11e7-8d9c-6b938b1543dd.png">
<img width="667" alt="screenshot 2017-06-30 08 34 18" src="https://user-images.githubusercontent.com/1711854/27735799-555420b2-5d6f-11e7-812b-73f92907f83e.png">

larrysalibra:
Great idea @guylepage3 !

larrysalibra:
For an initial implementation of this feature, we'd like to make it so that the Blockstack macOS app configures itself on first one to start automatically when the user logs into their macOS account. 

When quitting blockstack we should add a check box option that allows users to disable the start on log in functionality:

<img width="427" alt="screen shot 2017-08-11 at 12 32 51 am" src="https://user-images.githubusercontent.com/597182/29181178-c8120050-7e2c-11e7-8314-a4501e76f4ef.png">

We can modify the above dialog to include a checkbox option "Start Blockstack on login" in a similar to the dialog below:

<img width="427" alt="screen shot 2017-08-11 at 12 33 07 am" src="https://user-images.githubusercontent.com/597182/29181239-f790e3aa-7e2c-11e7-8520-b631ba929e3a.png">

When starting Blockstack on log in, the macOS app should **not** automatically open the user's web browser.

shea256:
> When starting Blockstack on log in, the macOS app should not automatically open the user's web browser.

Good call on this.

larrysalibra:
Thanks to @yknl this is complete and will be shipped in the next release!

guylepage3:
Awesome! Thanks @yknl 😄

PR #681, opened 2017-08-13
Add ability to start blockstack automatically on log in. 

yknl:
Automatically start Blockstack on log in with option to disable on quitting. #579

exit-prompt-start-on-log-in

Implemented using Service Manager `SMLoginItemSetEnabled` and a separate launcher service `BlockstackLauncher`. On startup, the launcher service runs and launches the main app without opening the web browser. Then the launcher service will terminate.

#### Testing notes:
Archive and "Export a Developer ID-signed Application"
App must be copied to Applications before running, or start on log in might not work.

#### Troubleshooting:
Sometimes Service Manager may get confused due to development copies of the main app and the launcher in xcode archives and derived data directory. Those might need to be cleared for this feature to work properly on developer machines.


larrysalibra:
Thank you @yknl ! This is really great. Tested it out this afternoon and it works perfectly for me.

> Sometimes Service Manager may get confused due to development copies of the main app and the launcher in xcode archives and derived data directory. Those might need to be cleared for this feature to work properly on developer machines.

I've noticed a similar with handling of `blockstack:` protocol -> macOS seems to get confused by multiple copies of the app on developer machines. Thanks for the heads up - it saved a lot of time testing it out.

Issue #670, opened 2017-08-04
Create new profile hangs after 2 wrong passwords


larrysalibra:
Entering password wrong twice leaves the create new profile modal in an unusable state:

<img width="449" alt="screen shot 2017-08-04 at 7 16 52 pm" src="https://user-images.githubusercontent.com/597182/28966573-8b8d8e02-7949-11e7-8383-871bcf43e4fa.png">

Thanks to @jackzampolin for reporting.

sieut:
Hi @larrysalibra, I'd love to work on this issue if you haven't started on it yet

sieut:
@larrysalibra I opened a PR for the issue https://github.com/blockstack/blockstack-browser/pull/682

larrysalibra:
Thanks for taking care of this @sieut ! Very happy to have you as a contributor!

PR #682, opened 2017-08-14
Fix for issue #670

sieut:


larrysalibra:
Thank you @sieut - this fixes the problem in #670

PR #727, opened 2017-09-05
Small docker launcher fixes

kantai:
Fixes a couple of issues in the docker launcher -- most notably, the `sed` command for changing bind addresses, setting the environment variable for the corsproxy, and issues when `$homedir` contains whitespace.


jackzampolin:
This looks good. Noticed that was a bit broken last week and didn't open an issue :( Thanks for picking it up! I'll go ahead and merge this.

PR #748, opened 2017-09-11
Docker launcher improvements, fixes, and windows support

kantai:
Adds:

* Password prompt for creating new wallet or just starting up without being given the password
* `BLOCKSTACK_DEBUG` flag with `lsof` installation and `--debug` execution
* Windows browser startup (required renaming start -> start-containers)

Fixes, seemingly a second time:

* `sed` errors in setting up the client.ini
* CORS proxy environment variables
* removes Windows warnings

Regarding the three fixes above -- those were fixed in PR #727 and merged into `develop`, but I guess they got rebased out?

### DIFF
--- a/launcher
+++ b/launcher
@@ -11,6 +11,14 @@ containerdir=/root/.blockstack
 corecontainer=blockstack-api
 # Name of Blockstack Browser container
 browsercontainer=blockstack-browser
+# Local temporary directory
+tmpdir=/tmp/.blockstack_tmp
+if [ ! -e $tmpdir ]; then
+   mkdir -p $tmpdir
+fi
+# set password blank so we know when to prompt
+password=0
+
 
 build () {
   echo "Building blockstack docker image. This might take a minute..."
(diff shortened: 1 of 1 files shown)


launcher
```diff
@@ -22,21 +30,34 @@ create-wallet () {
...
+  docker run -it -v "$homedir":$containerdir $coreimage sed -i 's/api_endpoint_host = localhost/api_endpoint_host = 0.0.0.0/' $containerdir/client.ini
 }
 
-start () {
+start-containers () {
   # Check for args first
-  if [ $# -eq 0 ]; then
-    echo "Need to input password for wallet located in the $HOME/.blockstack folder when staring api: ./launcher start mypass"
-    exit 1
+  if [ $# -ne 0 ]; then
```
jackzampolin: Thank you for fixing my hacky bash here!

launcher
```diff
@@ -22,21 +30,34 @@ create-wallet () {
...
+
   # Use init containers to set the API bind to 0.0.0.0
-  docker run -it -v $homedir:$containerdir $coreimage sed -i 's/api_endpoint_bind = localhost/api_endpoint_bind = 0.0.0.0/' $containerdir/client.in -e CORSPROXY_HOST="0.0.0.0"i
-  docker run -it -v $homedir:$containerdir $coreimage sed -i 's/api_endpoint_host = localhost/api_endpoint_host = 0.0.0.0/' $containerdir/client.ini
+  docker run -it -v "$homedir":$containerdir $coreimage sed -i 's/api_endpoint_bind = localhost/api_endpoint_bind = 0.0.0.0/' $containerdir/client.ini
+  docker run -it -v "$homedir":$containerdir $coreimage sed -i 's/api_endpoint_host = localhost/api_endpoint_host = 0.0.0.0/' $containerdir/client.ini
 }
 
-start () {
+start-containers () {
```
jackzampolin: Any reason for this name change? By convention I normally have it so `command` and `function` are the same:

```
command)
  function
  ;;
```

But doesn't have to be that way. Should we change the `command` to `start-containers` too?

launcher
```diff
@@ -22,21 +30,34 @@ create-wallet () {
...
   # Check for args first
-  if [ $# -eq 0 ]; then
-    echo "Need to input password for wallet located in the $HOME/.blockstack folder when staring api: ./launcher start mypass"
-    exit 1
+  if [ $# -ne 0 ]; then
+      password=$1
+  fi
+
+  # let's see if we should create a new wallet
```
jackzampolin: Way better way to implement. Like how you are not making user type PW in plaintext like I was 🙊.

launcher
```diff
@@ -11,6 +11,14 @@ containerdir=/root/.blockstack
 corecontainer=blockstack-api
 # Name of Blockstack Browser container
 browsercontainer=blockstack-browser
+# Local temporary directory
+tmpdir=/tmp/.blockstack_tmp
```
jackzampolin: Is this necessary? Doesn't adding the `lsof` install fix the issue that required this?

launcher
```diff
@@ -46,19 +67,29 @@ start () {
...
-      docker run -d --name $corecontainer -v $homedir:$containerdir -p 6270:6270 $coreimage blockstack api start-foreground --password $1 --api_password $1
-    elif [[ $(uname) == 'Windows' ]]; then
-      echo "Don't know if this works!!!"
-      docker run -d --name $corecontainer -v $homedir:$containerdir -p 6270:6270 $coreimage blockstack api start-foreground --password $1 --api_password $1
+      docker run -dt --name $corecontainer -v "$tmpdir":/tmp -v "$homedir":$containerdir -p 6270:6270 $coreimage $runcommand
+    elif [[ $(uname) == 'Windows' || $(uname) == 'MINGW64_NT-10.0' ]]; then
+      docker run -dt --name $corecontainer -v "$tmpdir":/tmp -v "$homedir":$containerdir -p 6270:6270 $coreimage $runcommand
+    fi
+
+    if [ "$BLOCKSTACK_DEBUG" == "1" ]; then
```
jackzampolin: This is a cool feature. Ideally we document this.

jackzampolin:
I think the rebase out is my fault 🙊. Sorry about that and thanks for fixing them again.

launcher
kantai: Yeah -- `start` is the Windows command name of something like macOS's `open`. If it's a function in the bash script, you can't call it. Of course, we could alias it, but this was simpler.

launcher
kantai: Oh yeah, if `lsof` is added to the install this isn't required, I'll remove

launcher
kantai: Yeah this `install lsof` line is unnecessary if it's been added to the dockerfile. Will remove.

kantai:
Closing this, since it seems to make sense to move `launcher` over to the packaging repo. Windows users and Linux users should be directed there (https://github.com/blockstack/packaging/tree/master/browser-core-docker#installing-blockstack-browser-and-api-with-docker) to setup a docker-ized browser.

Issue #770, opened 2017-09-14
Button and Links: On hover need to display pointer cursor


kkomaz:
At the moment, hovering over a button or a link showing a cursor is not consistent.  When a user hovers those elements, it should.

PR #771, opened 2017-09-14
Add cursor pointer to button and a tags

kkomaz:
Closes: https://github.com/blockstack/blockstack-browser/issues/770

- Add css styles `cursor: pointer` to both `<a>` and `<button>` tags


guylepage3:
@kkomaz this is awesome! 😄 Thanks again for the help. I will be merging things in over the weekend.

Issue #794, opened 2017-09-18
update tooltips


guylepage3:
Update tooltips to match design

<img width="502" alt="screenshot 2017-09-18 15 31 38" src="https://user-images.githubusercontent.com/1711854/30560637-8991ba46-9c86-11e7-8816-9a5d5a45eb09.png">

guylepage3:
@kkomaz want to update the tool tips to match the layout below?

![browser-macos-40b-profile-d16 2x](https://user-images.githubusercontent.com/1711854/30564895-78114dc8-9c94-11e7-97e9-0c1def9dc3e6.png)

Sketch file is located here: https://github.com/blockstack/designs/blob/master/browser/browser-macOS-40-profile-d16.sketch

kkomaz:
@guylepage3  Just so no one else starts on this...
https://github.com/blockstack/blockstack-browser/pull/800

guylepage3:
Thanks @kkomaz 👍

guylepage3:
complete. thanks again @kkomaz 😄

Issue #795, opened 2017-09-18
setup regtest


guylepage3:
Setup regtest

Follow instructions here: https://github.com/blockstack/blockstack-core/blob/master/docs/interactive_regtest_macos.md

- [x] Install dependencies
```
$ brew install automake berkeley-db4 libtool boost --c++11 miniupnpc openssl pkg-config protobuf qt libevent
```
- [x] Clone the bitcoin source code and cd into `bitcoin`
```
$ git clone https://github.com/bitcoin/bitcoin
$ cd bitcoin
```
- [x] Build bitcoin-core
```
$ ./autogen.sh
$ ./configure
$ make
```
- [x] It is recommended to build and run the unit tests
```
make check
```
- [ ] Run integration test

guylepage3:
on hold... was not able to get setup.

@kantai is building out a docker image for it.

guylepage3:
Ended up setting up a Docker image thanks to @kantai. 😄 

https://github.com/blockstack/blockstack-core/blob/develop/docs/interactive_regtest_macos.md#using-docker

guylepage3:
complete

PR #800, opened 2017-09-19
Tool-tip wrapper component create and linter fix

kkomaz:
Closes  https://github.com/blockstack/blockstack-browser/issues/794#issuecomment-330358289

- Created a new component called `ToolTip`.  Wrapper around `ReactToolTip` library component.  This is so we don't have to recreate the same component with the same configurations every time ToolTip is used.  The default position will on the bottom as opposed to the top.
- Also fixed linter warnings/error.

Components affected:
1. DefaultProfilePage.js
2. ViewProfilePage.js 

@guylepage3 I couldn't test it on `ViewProfilePage` because I didn't have an account name.  Can you confirm the positioning is correct?

screenshot of google chrome 9-18-17 6-31-06 pm

### DIFF
--- a/app/js/components/ToolTip.js
+++ b/app/js/components/ToolTip.js
@@ -1,0 +1,39 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ReactTooltip from 'react-tooltip'
+
+const ToolTip = (props) => (
+  <ReactTooltip
+    place={props.place}
+    type={props.type}
+    effect={props.effect}
+    id={props.id}
+    className={props.className}
+    offset={props.offset}
+  >
+    {props.children}
+  </ReactTooltip>
+)
+
+ToolTip.propTypes = {
+  place: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+  effect: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


guylepage3:
This is awesome @kkomaz! 😄 

For testing.. Let's get you setup with your own testing environment.. You'll need to install docker and `bitcoind`. 

Here are the instructions - https://github.com/blockstack/blockstack-browser/issues/795#issue-258595275

kkomaz:
@guylepage3 

Set up my test environment.  I realized based on the code that the tooltip should not show when user has a profile name.  Only applicable for non username profiles.

yknl:
This works well, I've merged it into develop. Thanks @kkomaz !

guylepage3:
Can we see a screen shot of the final iteration of this? Thank you sir. 👍

yknl:
image

Removed second line in tool tip as we are showing the owner address regardless if the user has a username

guylepage3:
I think this is close but not quite there. on items super small such as this, every pixel matters.. See tooltip design below..

browser-macos-40b-profile-d18 2x

guylepage3:
let's reopen this issue

guylepage3:
opening an issue for this.. here: https://github.com/blockstack/blockstack-browser/issues/904

PR #808, opened 2017-09-19
Add gaia hub support to browser

kantai:
This adds support for connecting to the gaiahub service to blockstack browser. This defaults currently to @jackzampolin's deployed version at 

```
https://hub.technofractal.com/hub_info
```

* It adds a configuration option to the settings pane
* It changes the is-storage-connected detection code to use a new variable `api.storageConnected`
* It generates the access token used to connect to gaiaHub and pushes the configuration settings to blockstack-core
* It writes the profile and profile image to the gaiaHub


kantai:
BTW -- to connect the gaia-hub to blockstack-core, you will need to be running blockstack-core@develop

kantai:
Oh no! The `authentication:` header should really be an `authorization:` header. Pushing update to gaia/hub, blockstack-core, and this branch on browser now.

larrysalibra:
Got this working on my machine! Great job @kantai & @jackzampolin!

Before I merge this in, I want to make sure I understand the requirements: a few questions:

* I noticed there are no more storage providers on the options page. are we only supporting gaia storage hubs going forward? (I would support this if we make it easy for others to run their own)
* If we're going to continue to support dropbox going forward, do you have any thoughts on how we should handle users trying to switch storage providers?

kantai:
* No more storage providers on the options page is a bug that I have introduced -- can fix that.
* If we're going to continue to support dropbox, users switching storage providers should (in theory) be pretty seamless -- when a user adds a new storage provider, browser should set the storage config options on core, and core will eventually replicate to the new provider (I believe through using the local disk replica)

larrysalibra:
> No more storage providers on the options page is a bug that I have introduced -- can fix that.
No need to fix - we can do that - just wanted to confirm whether or not it was intentional.

> If we're going to continue to support dropbox, users switching storage providers should (in theory) be pretty seamless -- when a user adds a new storage provider, browser should set the storage config options on core, and core will eventually replicate to the new provider (I believe through using the local disk replica)

Sounds like in practice this has the potential to raise lots of user questions about what's happening.

At some point, if the user is disconnecting the original storage provider, they'd also have to spend bitcoin sending a name update with the new zone file pointing to the new storage provider.

Given our tight time frame, I'd tend to want to impose artificial constraints on what the user can do to minimize complexity.

If we feel it makes sense to continue supporting dropbox for the time being so we have multiple storage providers, then we could dissuade people from switch providers once you have a blockchain ID with a profile stored on one or the other and implement it in the future.

Another option is to drop dropbox support (ha! drop!) and have an advanced setting in the storage provider window that lets you point your browser to your own gaia hub. This avoids the switching problem while still having decentralized options, but only for people comfortable getting their hands dirty.

We could also consider adding a coming soon option to point it to your Casa home node (not sure exactly what they're calling it). 

---

It looks like it's also currently hardcoded to use the private key of the first owner address.

We'll want writes for each separate ID to use the ID's respective owner address private key, right?

jackzampolin:
FWIW I've tried to make the Gaia Hub very easy to deploy. Have a [full writeup on how to run your own](https://github.com/blockstack/gaia/tree/master/hub) in the README. The only thing I could do to make it easier is to dockerize the nginx instance and have some notes about firewalls on different cloud services. I'm also pro keeping dropbox for the time being.

kantai:
> If we feel it makes sense to continue supporting dropbox for the time being so we have multiple storage providers, then we could dissuade people from switch providers once you have a blockchain ID with a profile stored on one or the other and implement it in the future.

I agree with that approach. We would also need a note on the "restore account" screen that the user should chose to reconnect with the same storage provider. (Alternatively, browser could read their zonefile and try to figure out their storage provider -- however, that would require a different approach to the restore code paths in browser I think?)

Another note re: gaia-hub support. This PR doesn't include the social proof header, I'm working on that today -- it will require support in both core and browser, as both of them submit storage requests, and we'll include the social proof header in all requests.

larrysalibra:
From slack:

larry [12:17 AM] 
@kantai one of the reasons we wanted to add gaia stores was to simplify the onboarding process


[12:17] 
storage hubs


[12:18] 
by removing the connect your XXXX step


[12:18] 
so we can make it use gaia hubs by default


[12:18] 
and skip that step


[12:18] 
but that leads to the question of how to give users an alternate choice


[12:18] 
didn't know if you had any thoughts about that


aaronb [12:19 AM] 
I have 2 sort of branches of thought


[12:19] 
first -- the gaia hub is really easy to deploy, and you can configure it as your "personal storage provider" backed with whatever store you want


larry [12:20 AM] 
it seems like avoiding the redirect may be enough of an improvement - and we could keep the pick a provider on-boarding page. that way users would have some idea where their data is being stored and understand that they have a choice. the downside is that a lot of people will probably choose dropbox because it is a familiar brand.


[12:21] 
when given a choice, i usually pick the one I have and recognize - but we don't to send people into that redirect to dropbox unless they realllllly want it


aaronb [12:21 AM] 
yeah, I had originally thought we'd keep the storage provider picker in the on-boarding


[12:22] 
I'm not sure which way most people would go when picking the storage


[12:23] 
I'm always inclined to not link an account if I don't have to


larry [12:23 AM] 
so we can make it so that gaia hub appears as already selected and the default, ideal option


aaronb [12:23 AM] 
yeah


larry [12:23 AM] 
and de-emphasize dropbox


[12:24] 
somehow communicate that it is harder and you really dont want to do that :-)


[12:24] 
okay cool


[12:24] 
the other thing i was thinking


[12:24] 
was to totally get rid of it


[12:24] 
use gaia hubs by default


aaronb [12:24 AM] 
"Connect to Dropbox (slow)"



larry [12:25 AM] 
and then after onboarding have a dismissible tip or something that says "want to use your own storage? click here"


aaronb [12:25 AM] 
ah


[12:25] 
yeah, I actually like that a lot


[12:25] 
as long as the tooltip pops up


larry [12:25 AM] 
OR!


aaronb [12:25 AM] 
and says "we connected you to a Blockstack provided storage option. want to use your own?"


larry [12:26 AM] 
we could have a storage screen that's just informational


[12:26] 
"your browser already comes configured to use a blockstack PBC provided gaia hub - to connect your storage click here"


[12:26] 
which shows the actual "connect dropbox" page


aaronb [12:27 AM] 
yeah, either way -- I think it works to do the default automatically, so long as we get the information across to the user


larry [12:27 AM] 
ok cool


aaronb [12:27 AM] 
and the storage provider page could also have an option for selecting your own hub


larry [12:27 AM] 
what information would the user need to provide for their own hub?


[12:27] 
a url and?


aaronb [12:27 AM] 
just a url


larry [12:28 AM] 
perfect

aaronb [12:29 AM] 
yeah, I think if people are going to run their own hubs, we should add a "whitelist addresses" config option in the hub


[12:29] 
so that your hub isn't providing storage to everyone :slightly_smiling_face:


larry [12:29 AM] 
yeah - security through obscurity is enough now - no one will look at the zone file lol


[12:29] 
#FamousLastWords


aaronb [12:29 AM] 
it doesn't even show up in the zonefile!


[12:29] 
you get a public url


larry [12:29 AM] 
oh right! the endpoint could be anywhere!


[12:30] 
that's why i was seeing windows.net and technofractal in my urls


aaronb [12:30 AM] 
yeah, my local one doesn't bind to 0.0.0.0


[12:30] 
so it already won't accept other people's storage requests. but it won't accept mine if I was on another device


larry [12:31 AM] 
beautiful


[12:31] 
this is really elegant - excited about this.


[12:32] 
so my last question: should i hold off on integrating this stuff until the social proofs stuff is ready?


[12:32] 
what's the time frame for that


[12:32] 
oh and is that going to be really hard to test in regtest?


aaronb [12:32 AM] 
well, I was going to merge the social proof support into the gaia-hub today


[12:33] 
it works with it's checks -- but we need to figure out how we want them actually passed from the browser during initialization


larry [12:33 AM] 
it can't pass them until the user has made them


aaronb [12:33 AM] 
yeah, exactly


[12:34] 
so it may have to be an off-line thing


larry [12:34 AM] 
off-line?


aaronb [12:34 AM] 
oh, outside-of-the-write-path


larry [12:34 AM] 
ahh - like any user who is unverified can write a profile file


[12:34] 
with a different endpoint


aaronb [12:35 AM] 
I was thinking more along the lines of -- a proof is submitted when they're made in the client


[12:36] 
and the storage service allows you to perform writes for a little while before you prove yourself


[12:36] 
but, yeah, writing the profile file unverified works as well (edited)


[12:36] 
and then you just check the proofs in the profile


larry [12:37 AM] 
so we'll have to make users go through this verification process for each ID they create


[12:37] 
even if the 0th one is verified


[12:37] 
and they create the 1st one


[12:37] 
they'll need to add proofs for it


[12:37] 
is that understanding correct?


aaronb [12:37 AM] 
the storage service uses the 0th profile's address for verifying writes


[12:38] 
so, only the 0th profile needs to be verified


ken [12:38 AM] 
is the 0th profile the same as the default profile?


larry [12:38 AM] 
yeah


[12:39] 
wait no


aaronb [12:39 AM] 
not currently :neutral_face:


larry [12:39 AM] 
the 0th is the first one


[12:39] 
the owner addresses are derived by an index starting at 0


[12:39] 
the default one is which ever one the user selects as default


[12:39] 
which is by default the 0th one when there is only one


[12:39] 
:-D


ken [12:39 AM] 
right


[12:40] 
as a user that could be a point of confusion

[12:40] 
if I have a default profile that I verified, but it’s not the 0th one

larry [12:40 AM] 
yeah i think that's a bit weird too

aaronb [12:40 AM] 
yeah -- we could support using the default one


larry [12:40 AM] 
i would have thought that each profile that wants to write needs to be verified

[12:41] 
which would also be annoying

aaronb [12:41 AM] 
yeah, that's a bit of a vestigial feature

[12:42] 
the way core looks at storage is you have a storage provider for the core node

larry [12:42 AM] 
oh right we're doing this through core still

ken [12:44 AM] 
should we add verifications to the onboarding flow?

[12:45] 
skippable of course

larry [12:47 AM] 
yeah i think we need to

[12:47] 
i think the default profile would also be confusing

[12:48] 
people will think "the the storage was working, why doesn't it work now after i switched default profiles" (edited)

ken [12:49 AM] 
we could ask users to connect storage separately for each profile they create

[12:50] 
I don’t know how feasible that is with the current core setup though

larry [12:50 AM] 
lots of changes

[12:50] 
@kantai how feasible is it for any verified profile to work?

[12:50] 
with a reasonable limit on index

[12:51] 
like say any verified profile from 0 to 10 or 20?

aaronb [12:51 AM] 
Pretty feasible

[12:51] 
We could have it search profiles

larry [12:52 AM] 
the hub right?

aaronb [12:52 AM] 
Or have it pass it in a header

larry [12:54 AM] 
so from the user's perspective, as long as any profile is verified, then it will work with out regard to it being default or not.

[12:54] 
and if there's not at least one verified profile then browser can prompt the user to verify one

[12:55] 
or core can return a specific error for htat case and we can handle it and display it to the user

aaronb [12:58 AM] 
Yeah -- I'll write up how I think this would work this afternoon

siddhant [1:00 AM] 
joined #engineering.

jack [1:25 AM] 
Just catching up on this. Great discussion

aaronb [1:29 AM] 
@larrysalibra how do the social proofs work *before* a profile has a domain name? (edited)

[1:29] 
because for now, browser doesn't store the profile until a name has been associated with it, right?

larry [1:31 AM] 
@kantai right - I need to change it so that it stores a profile before it has a name.

[1:33] 
A lot of the browser was designed with the assumption of a username - we changed that right before the consensus announcement. Opened an issue about that -> https://github.com/blockstack/blockstack-browser/issues/822 (edited)


aaronb [1:33 AM] 
ah gotcha

[1:35] 
okay, let me talk through an example to make sure we're on the same page --

[1:35] 
I have a keychain on browser

[1:35] 
0'th profile has address -- 1EkbYpQES4SoxbtSrasApaZqHgS5k2r2UG

[1:36] 
1'st profile has address -- 1Pa7JiqFvTjGx9EeyMzd73zqSAKwFymG15

[1:36] 
when I connect to gaia-hub, I connect with my default profile

[1:37] 
that uses addresses on gaia-hub like `/store/1EkbYpQES4SoxbtSrasApaZqHgS5k2r2UG/foo.profile.json`

[1:37] 
and since storage is per-core-node rather than per-profile, 1'st profile uses addresses like `/store/1EkbYpQES4SoxbtSrasApaZqHgS5k2r2UG/bar.profile.json`

larry [1:38 AM] 
i was thinking we could come up with a convention for the location of the profile that's name independent something like `/store/1EkbYpQES4SoxbtSrasApaZqHgS5k2r2UG/keyfile.json`

aaronb [1:39 AM] 
hmm

larry [1:39 AM] 
where "keyfile" is whatever we're calling the profile/token/keyfile :slightly_smiling_face:

aaronb [1:39 AM] 
yeah

[1:39] 
I'm just trying to think through how we could store things for profile 1 in `/store/1Pa7JiqFvTjGx9EeyMzd73zqSAKwFymG15`


[1:40] 
and things for profile 0 in `/store/1EkbYpQES4SoxbtSrasApaZqHgS5k2r2UG`


larry [1:40 AM] 
i assumed we'd sign requests with the other key


[1:40] 
i was thinking more out into the future where this is all in blockstack.js and not in core


aaronb [1:41 AM] 
yeah, I think that'd be more doable


[1:41] 
I might be able to make the core driver differentiate based on the address of the writer


larry [1:42 AM] 
the reason i used the name to namespace the data in dropbox, was at the time, the name was seen as you ultimate identity, not the owner address...so if you transferred the name, you'd still want to have access to the same data (edited)


[1:42] 
that was at least my thinking at the time


aaronb [1:43 AM] 
ah


larry [1:43 AM] 
all of this time i was thinking i was `larry.id` when it turns out i am actually really `15kxt5aboFnBuJseR6cHNxMG7Tn6tfkrQZ`



aaronb [1:58 AM] 
okay -- so basically, the issue we have, is that the core driver doesn't "know about" the address it's storing on behalf of


[1:59] 
that could be changed, but it could touch lots of other things


larry [2:03 AM] 
can we pass the new setting in with each write request?


aaronb [2:04 AM] 
yeah -- I think it could work


[2:04] 
blockstack-storage.js would have to add the new setting


[2:05] 
`address : 15kxt5aboFnBuJseR6cHNxMG7Tn6tfkrQZ` and `signedChallenge : <hex-string>`


[2:06] 
and then core would only be a relay


[2:06] 
but at that point it could even just do the operation in javascript


larry [2:07 AM] 
doing it only in javascript would be ideal - how far are we from that in the current implementation?


[2:08] 
i recall jude saying something about implementing gaia in javascript

kantai:
Summary -- there are basically 2 problems that are outlined above.

1. This one is kind of fundamental -- gaia-hub's authenticate writes to on behalf of an _address_ with a signed challenge. blockstack-core's storage strategy doesn't relay enough information to the storage driver to change the write address/authorization header based on who the write is performed on behalf of.
2. Social proofs are required for writes, but those would need to be written to the profile first. This is true if the social proofs are verified by reading the profile/keyfile -- and solvable by allowing unverified profile writes. However, if the social proof array is passed directly as a header, then that's not an issue.


Problem 1 is solvable through a variety of hacks or re-designs:
1. Punt on the problem, allow all the profiles to write with the same _address_ and signed challenge. This is what is currently implemented.
2. Forward the information about the _address_ and the signed challenge to core from `blockstack-storage.js` -- this would require adding the signed challenge to the session token.
3. If we have the _address_ and the signed challenge available to the app _anyways_, we could just perform the write directly from javascript. (PROBLEM: writes no longer propagate to local disk)

(3) would basically be a javascript library for interacting with a "gaia hub", rather than blockstack-core, but would need to be writing compatibly for multi-user storage to work correctly. Without changing the protocol and assumptions being made in the javascript, (3) has little benefit over (2).

(2) and (3) might require a synchronized update of blockstack.js and core, because the authtoken now includes new things plus would require passing new query parameters to core.

(1) is inelegant and kludgey, but it works and isn't really strange until people have multiple profiles.

This is getting kind of long, but if we're likely changing assumptions about how storage should behave (and moving to a more "identity/user storage hub" model), then that means we shouldn't be trying to find a perfect solution today, and instead be happy with a kludgey solution.

jackzampolin:
```This is getting kind of long, but if we're likely changing assumptions about how storage should behave (and moving to a more "identity/user storage hub" model), then that means we shouldn't be trying to find a perfect solution today, and instead be happy with a kludgey solution.```

I think this is the salient point here. This is going to be the first implementation. Also the usecase is going to be different in the future (individual users, not anyone who wants). I think support for multiple profiles is pretty necessary however so I would be biased toward 2. I also think support for restoring multiple profiles is a feature we need to have.

kantai:
The current setup _does_ support multiple profiles, it just has some strange consequences if you try to switch which profile you authenticate to storage with. Think of it like the Dropbox driver -- the same dropbox driver is used for all of your profiles -- but with the gaia-hub driver, you authenticate with a address+signature instead of a Dropbox account. I agree, multiple profiles all using the same address+signature isn't ideal, and it's especially not good when you are expecting the main profile to have social proofs.

This reminded me... passing the address+signature to the app is the equivalent of passing your storage credentials to an application. That is no good (you'd want to 'scope' the paths that the application could write to). Core could instead get the credential during the authentication phase, and not forward it to the application, then core would track different credentials for different app sessions.

larrysalibra:
+1 to passing social proofs in header
+1 to going forward with the current kludgey solution in the short term.

Just wanted to point out that storing data for multiple profiles until address+signature of one profile is is flawed from a privacy perspective. It makes it very clear that all of these supposedly independent profiles are actually controlled by the same person and part of the same keychain. 

We don't have this problem to the same extent with the dropbox driver. Dropbox knows the the multiple IDs are controlled by the same person because they are stored in the same dropbox account, but this isn't apparent unless you are Dropbox.

As a proof of concept, kludgey solution, this is probably fine. It's not for a real solution.

--

> This reminded me... passing the address+signature to the app is the equivalent of passing your storage credentials to an application. That is no good (you'd want to 'scope' the paths that the application could write to). Core could instead get the credential during the authentication phase, and not forward it to the application, then core would track different credentials for different app sessions.

Given the presence of a key file, you wouldn't need to pass the ID address+signature to the app. The app could use its own address (generated from app key that the browser sends to the app as part of sign in) and signature to make writes. You'd need to a mechanism for the storage node to check that the app key is actually in the key file of the ID with valid write ability (proofs, payment or whatever the node requires).

Something to think about for the future.

kantai:
>Just wanted to point out that storing data for multiple profiles until address+signature of one profile is is flawed from a privacy perspective. It makes it very clear that all of these supposedly independent profiles are actually controlled by the same person and part of the same keychain.

Yep -- I agree with this -- and once we move towards a more hub-centric storage model, we'd be able to implement this pretty smoothly.

> Given the presence of a key file, you wouldn't need to pass the ID address+signature to the app. The app could use its own address (generated from app key that the browser sends to the app as part of sign in) and signature to make writes. You'd need to a mechanism for the storage node to check that the app key is actually in the key file of the ID with valid write ability (proofs, payment or whatever the node requires).

Yes, we could use the keyfile for this. Could also be implemented without the keyfile by having the signed challenge text indicate an allowed path like "/blockstack-todos/". This would indicate that the app could only write to sub-paths of that path. This would be nice for collections, because you can implement that by simply using the same root of the paths. This would also limit the amount of keyfile lookups performed by the hub.

kantai:
Closing this, as support has moved over to #833

larrysalibra:
This is merged into `develop` thanks again @kantai & @jackzampolin for getting gaia hubs up and running and the rest of the team for feedback on the high-level approach to storage providers!

Issue #622, opened 2017-07-14
Click/Tap to verify identity key


larrysalibra:
Switch to click to verify identity key similar to copay:

![28137712-c4ea0bce-6780-11e7-9022-118dfe376e13](https://user-images.githubusercontent.com/597182/28218436-19ad9c1e-68eb-11e7-8575-7f95474d1558.PNG)

PR #809, opened 2017-09-20
initial concept for #622

daviddahl:
This is an initial concept for the "tap to verify passphrase words" #622 

No doubt there are some style issues and there are no tests written yet, but I thought I would submit this after a couple hours of hacking. 

I tried to change as little as possible, even leaving the full submission form (but hidden). I did not dig in enough yet to see what you are doing to optionally add classes to nodes, etc. I just wanted to get the functionality working.

A question about interaction: Do you want to display any kind of warning when the wrong word is tapped on? For now, it just does nothing, which is clear to me as a nerd that I tapped in the wrong order:)


daviddahl:
git hath no fury...

Issue #802, opened 2017-09-19
Search for username dialog button color 


larrysalibra:
Looks weird with a white button.
<img width="648" alt="screen shot 2017-09-19 at 6 58 19 pm" src="https://user-images.githubusercontent.com/597182/30589443-06fc1396-9d6d-11e7-8afe-eb2df4c3549d.png">

kkomaz:
I can take a look at this @guylepage3   Let me know what color you're thinking.

guylepage3:
@kkomaz sounds great... as a general rule of thumb we use `btn-primary btn-block` for our buttons.

larrysalibra:
This was fixed.

PR #823, opened 2017-09-21
Fix Search Button styles and refactor button vs. link logic

kkomaz:
Fixes: https://github.com/blockstack/blockstack-browser/issues/802

1. Per discussion with @guylepage3, I moved the cancel button inside of each component that utilizes the cancel button
2. Made the components into class components.

@yknl I noticed we use the `react-router` `<Link />` component but style them with `btn btn-primary btn-block`.  I think if the purpose of the component is to be a **button** we should wrap it with a `button` element and style them appropriately with the classNames.  To fix the issue of navigating away we can use the `browserHistory` functionality and push those routes `onClick` as needed.  What do you think?

demo

### DIFF
--- a/app/js/profiles/components/registration/RegistrationSearchBox.js
+++ b/app/js/profiles/components/registration/RegistrationSearchBox.js
@@ -1,50 +1,74 @@
-import React, { PropTypes } from 'react'
+import React, { PropTypes, Component } from 'react'
 import Alert from '../../../components/Alert'
+import { browserHistory } from 'react-router'
 
-const RegistrationSearchBox = (props) =>
-  (
-  <div>
-    <h3 className="modal-heading">Search for your username</h3>
-    <div className="modal-body">
-    {
-      props.alerts.map((alert, index) =>
-         (
-        <Alert
-          key={index} message={alert.message} status={alert.status} url={alert.url}
-        />
-         )
-      )
-    }
-      <p>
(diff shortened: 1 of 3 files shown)


guylepage3:
this is awesome @kkomaz!!!! 😄

guylepage3:
@kkomaz I see what you're suggesting with `browserHistory` but we should keep consistent for now with the React Router `<Link />` component.

Can you make that change?

kkomaz:
@guylepage3 

One of the issues I noticed with using `<Link>` is that the CSS doesn't play well with it.

`btn-block` spacing only applies if 

1. They are both buttons.
2. They are siblings.

screenshot of screenfloat 9-21-17 6-58-40 pm


Unfortunately one is a `Link` and the other is a `button` 😞   What do you think?

kkomaz:
I'll revert it back and see if I can make that spacing in CSS

kkomaz:
There was no css changes needed.  I just needed the links to be siblings. 

screenshot of google chrome 9-21-17 9-32-35 pm

screenshot of google chrome 9-21-17 9-32-58 pm

larrysalibra:
Looks good to to me! Thanks @kkomaz!

guylepage3:
@kkomaz there is probably the `form-group` or `form-control` container on the first modal. simply apply a container to the other modal windows to match the width of the `form-group`/`form-control`.

Issue #825, opened 2017-09-22
Absolute path support for imports


kkomaz:
From a development perspective, support for absolute path would be great.  With absolute path, we would no longer have to do deep nesting when importing utility functions or separate components.  Example below

```
import { DROPBOX } from '../../../account/utils/index'
```
to
```
import { DROPBOX } from 'js/account/utils/index'
```

kkomaz:
@guylepage3 This is an enhancement I can work on if you like the idea.

larrysalibra:
@kkomaz is this something that's possible?

kkomaz:
Yep!  Just need some minor configurations via
https://github.com/tleunen/babel-plugin-module-resolver

yknl:
This would be nice to have!

guylepage3:
Nice @kkomaz.. 😄

larrysalibra:
Would be nice to have. 

Keep in mind eslint and flow need to continue working along with the respective plugins in our editors. It looks like getting all of these to work requires adding a bunch of additional dependencies both to the project and to the plugins our editors use - not sure how well they are maintained.

I'm in general hesitant to add too many dependencies that aren't widely used, but feel free to give it a go.

If it works well happy to merge it.

dashaxiong:
It looks like this issue should be closed, as per issue #828 and @aulneau commit 403ea67b697d

wbobeirne:
^ Indeed!

PR #828, opened 2017-09-25
Absolute Path support via Babel resolver

kkomaz:
Closes: https://github.com/blockstack/blockstack-browser/issues/825

- Absolute path import support
- 2 Libraries added: babel-plugin-module-resolver && eslint-import-resolver-babel-module
- babel-plugin-module-resolver is responsible for the proper aliasing
- eslint-import-resolver-babel-module will fix the eslint issues
- Did some testing with several components to make sure nothing was breaking

To add additional aliasing it is as easy as going into the `.babelrc` and adding more keys to the "alias" object.  For example

```
"alias": {
  "js": "./app/js",
  "styles": "./app/styles"
}
```


larrysalibra:
Thanks @kkomaz - @cwackerfuss can you take a look at this? Here was our earlier discussion about it: https://github.com/blockstack/blockstack-browser/issues/825#issuecomment-331442929

aulneau:
This is great! I actually recently just implemented this here -> https://github.com/blockstack/blockstack-browser/commit/403ea67b697d889c91b5491a565eff29ad00732b without seeing this PR.

I think this would be a great addition! Most of the projects I work on/have worked on use this. Regarding your concerns @larrysalibra, the plugin has 156,038 downloads in the last week, so a lot of people use it and is well tested. 

I'd make the recommendation that we prefix our paths with `@` so there is a clear distinction between internal and external resources. cc @cwackerfuss

kkomaz:
Either way works for me.  This is a pretty old PR from awhile back.  @cwackerfuss

cwackerfuss:
Thanks, sorry about the confusion @kkomaz -- closing this now. @aulneau make sure to reference the issue in a PR once you push.

PR #840, opened 2017-10-02
remove setting host to 0.0.0.0

### LINKED ISSUE #838: 0.0.0.0 is an invalid address
image

After signing in to the Blockstack Todo App, no tasks can be saved and the console reports the above errors.

I've already done some preliminary investigations with the help of @kantai

xmakina:
Resolves #838


xmakina:
I don't have access to any other operating systems, but this has resolved the issue on my Windows machine

kantai:
Hi @xmakina, thanks for the PR!

Can you open the PR over here instead:

https://github.com/blockstack/packaging/blob/master/browser-core-docker/launcher

That's the version we include in the Windows install (and we plan to eventually remove the `launcher` file from the browser repo).

larrysalibra:
Thanks @xmakina

PR #872, opened 2017-10-08
Add section about app development

friedger:
Contains also copy cat from https://github.com/blockstack/blockstack-todos


larrysalibra:
Beautiful! Thank you @friedger!

Issue #901, opened 2017-10-12
Profile doesn't save when gaia hub connected


kantai:
The EditProfilePage checks for a dropbox api token before trying to save a user's profile. So if you connect a gaia hub, it won't save the profile.json object.

Fix is in `hotfix/connected-storage`, and I'll open PR.

larrysalibra:
This should be fixed in develop now. Please re-open if it happens again.

PR #902, opened 2017-10-12
Edit profile should check storageConnected rather than dropboxAccessToken

kantai:
This led to the bug in #901 -- the profile skips saves when connected via gaia hub instead of dropbox.


larrysalibra:
This is already fixed in a branch I’m working on.

larrysalibra:
Unless you want to release a new browser with this fix

kantai:
Nope, this can wait until the release. Closing.

PR #956, opened 2017-10-25
Removed the "portal" terminology from README

muneeb-ali:


muneeb-ali:
@larrysalibra this looks ready to merge, let me know if you have any comments.

larrysalibra:
Looks good. We should also remove the section about using docker to develop apps. Will merge this first.

Issue #958, opened 2017-10-25
Add transit encryption to application session JWT


kantai:
There's private information encoded in the JWT, which _should_ be encrypted using a transit key.

This shouldn't require any changes on the `blockstack-core` side (the jsonschema for the app session request should remain the same), but from the looks of it, will require updates of both the browser and `blockstack.js`

larrysalibra:
Here's the issue for blockstack.js https://github.com/blockstack/blockstack.js/issues/276

kantai:
This is implemented in the `feature/transit-encryption` branches of `blockstack.js` and `blockstack-browser`

It performs a hardcoded version check on the authentication request's payload. I've tested this with both browser and the todos application using the new `blockstack.js` -- I still need to test the other two configurations:

1. Browser with new features, auth request from old todos
2. Old browser, auth request from new todos

kantai:
Tested with the three relevant configurations now -- authentication and storage work in all the cases.

dominictarr:
there is also coreSession tokens - https://github.com/blockstack/blockstack.js/blob/master/src/auth/authSession.js#L35 this should really be done the same way as this, although having core never see the keys and just broadcast a transaction created in the app would be even better.

PR #960, opened 2017-10-25
Use transit key for encrypting private data in JWTs

kantai:
Discussion in Issue #958 

Right now, this uses hard-coded version checks on the JWT payload -- ultimately, we'll want a better versioning scheme, but this works for now, because there's only two versions.


larrysalibra:
Works for me! Thanks @kantai

PR #1039, opened 2017-11-07
Added password strength indicator to Welcome pages

### LINKED ISSUE #919: Add password strength indicator
Use this: https://github.com/dropbox/zxcvbn

jdlee022:
resolves #919 

Please let me know if there's anything I need to fix or change.

### DIFF
--- a/app/js/welcome/components/EnterPasswordView.js
+++ b/app/js/welcome/components/EnterPasswordView.js
@@ -3,6 +3,7 @@ import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
 import InputGroup from '../../components/InputGroup'
 import { AccountActions } from '../../account/store/account'
+import zxcvbn from 'zxcvbn'
 
 
 import log4js from 'log4js'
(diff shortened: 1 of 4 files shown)


maxstoller:
Hi @jdlee022, nice one.

I'm just getting acquainted with the codebase, but here's my feedback: You might want to consider adding the same strength indicator to the [ChangePasswordPage.js](https://github.com/blockstack/blockstack-browser/blob/4ac5c605ffdedaa1be68c92048fdadd70bda2b10/app/js/account/ChangePasswordPage.js) as well. It has its own logic around password strength ([L#70](https://github.com/blockstack/blockstack-browser/blob/4ac5c605ffdedaa1be68c92048fdadd70bda2b10/app/js/account/ChangePasswordPage.js#L70)). The password policies should be consistent across both pages. You could generalize the EnterPasswordView component and use it in both places.

larrysalibra:
@jdlee022 thanks so much for this. I modified it so that we only warn of weak passwords instead of preventing users from submitting the form with no feedback.

@maxstoller, your feedback is spot on. Want to open issues for these suggestions so that we can address them in a future iteration? We'll be cleaning up the on-boarding process in the near future which would be a great time to address these changes.

PR #1104, opened 2017-11-28
Light-weight multi-reader storage

yknl:
Adds ability to add custom key/values to the user profile. This enables a very light-weight form of multi-reader storage. For now custom key/values will need to be manually entered. An update to blockstack.js is planned to give apps the ability to programmatically add/edit key values with the user's permission. Also included in this PR is the ability to delete accounts.

image

Custom account in profile:
image

In edit mode, you can now click on the trash can to delete an account
image

### DIFF
--- a/app/js/profiles/DefaultProfilePage.js
+++ b/app/js/profiles/DefaultProfilePage.js
@@ -10,6 +10,7 @@ import { AccountActions }  from '../account/store/account'
 import SecondaryNavBar from '../components/SecondaryNavBar'
 import SocialAccountItem from './components/SocialAccountItem'
 import PGPAccountItem from './components/PGPAccountItem'
+import CustomAccountItem from './components/CustomAccountItem'
 import ProfileCompletion from './components/ProfileCompletion'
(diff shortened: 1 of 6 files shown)


larrysalibra:
Wow this looks great @yknl! Can’t wait to give it a try! Thanks!

guylepage3:
Very cool @yknl

larrysalibra:
@yknl sorry for taking so long to give feedback on this:

It works pretty well. 
Couple of things:

1. If i click on a custom entry and then try to change it, i end up getting a new one. I'd expect that it lets me edit the value.
2. There doesn't seem to be a way to delete an entry. We could add a delete button in the edit dialog.

Let me know what you think.

larrysalibra:
@yknl If I recall correctly, this was a different implementation of multi-reader storage that let us add custom accounts. Do we want to keep this functionality?

vsund:
This might be not relevant for multi-reader storage anymore, but I'd really like the ability to add custom accounts :)

yknl:
This PR is now obsolete

PR #1105, opened 2017-11-29
[Issue 866] - Add Windows Installation Instructions from packaging to browser README

frankchen07:
Moved the Windows Docker installation instructions from the [packaging repository](https://github.com/blockstack/packaging/tree/master/browser-core-docker) to the master Blockstack Browser README. 

I didn't find a README in the app folder of master as per the instructions, so I believe the one I updated should be it. 

It seems the Windows install instructions in the packaging repository are also present on the [Blockstack Core Installation README](https://github.com/blockstack/blockstack-core#installing-blockstack). 

I've added installation instructions to the Blockstack Browser README, but feel that the instructions utilizing the launcher script and Docker CLI should only be on the Blockstack Core README. 

I would suggest separating installation / initialization instructions explicitly organized by core / browser / etc (e.g keep the windows launcher instructions on the core README, and launcher commands for the browser on the browser README), but my understanding could be elementary, so let me know if that logic would be unwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,23 @@ Blockstack for macOS contains a Blockstack Core API endpoint & a CORS proxy.
 *Note: npm dev runs a BrowserSync process that watches the assets in `/app`, then builds them and places them in `/build`, and in turn serves them up on port 3000. When changes are made to the original files, they are rebuilt and re-synced to the browser frames you have open.*
 
 
+### Windows
+
+*Please note these instructions only apply for Windows 10.*
+
+The Blockstack API and the Blockstack Browser run best in Docker. There is a provided CLI to help you build and launch `docker` images if you are not comfortable with `docker`:`launcher`. The CLI will pull down the images from our [Quay image repository](https://quay.io/organization/blockstack).
+
+1. Download the [launcher script](https://raw.githubusercontent.com/blockstack/packaging/master/browser-core-docker/launcher) from our packaging repository.
+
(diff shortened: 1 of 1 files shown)


larrysalibra:
Thanks so much for this @frankchen07 !

@jackzampolin - I'm not up to do that on the latest windows situation. Can you take a look at this pull request?

jackzampolin:
So the issue here is that the windows build we ship does not include the API or the docker install. It is the JS only version of the browser. And that is where we are pointing users. If we want to merge (these are good, working instructions), I think we need to add an experimental or other note.

Another thing to consider is that we are phasing out the core API in a coming release and these instructions will be outdated and will need to be removed at that time. I'm unsure if this will be useful or confusing to users. I'm leaning toward confusing.

README.md
```diff
@@ -62,6 +62,23 @@ Blockstack for macOS contains a Blockstack Core API endpoint & a CORS proxy.
 *Note: npm dev runs a BrowserSync process that watches the assets in `/app`, then builds them and places them in `/build`, and in turn serves them up on port 3000. When changes are made to the original files, they are rebuilt and re-synced to the browser frames you have open.*
 
 
+### Windows
+
+*Please note these instructions only apply for Windows 10.*
```
jackzampolin: Please add a note here that this is for developers only. This note should also point users to our [download page](https://blockstack.org/install). If a user would like to install and try out Blockstack we have a build of the Browser that does not include the Core API. 

> Note: This installation is for setting up a development environment on Windows. If you are looking to try out, or use Blockstack for the first time please go to our downloads page and install the version there: blockstack.org/install

frankchen07:
@larrysalibra @jackzampolin - thanks for the review. I've added the development note, reworded some of the instructions for better flow and added a downloads page link to the specific area in the README.

jackzampolin:
Thanks @frankchen07! I'm going to go ahead and merge this!

PR #1122, opened 2017-12-17
app/js/account/utils/blockstack-inc.js

ivegotwings:
uploadPhotoToGaiaHub-> set proper image mime type


wbobeirne:
Sorry for leaving this unaddressed for a long time, it's gone stale so I've rewritten and PR'd with #1564. Thanks for the contribution though!

PR #1128, opened 2017-12-18
Skip Core Session Initialization for Applications Using Latest blockstack.js

kantai:
This PR does two things:

1. Checks an authRequest to see if it is coming from an application which can communicate with the gaia hub directly, and if so, does not send a core session token.
2. Removes the tooltip from applications which require storage on Windows and the Web app. If an authRequest comes in from an old `blockstack.js` and requests the `store` permission, the AuthModal displays a message saying that the app is using an old version of blockstack.js, and Windows users will get support once the application updates its library.

I believe the release schedule is to hold off on releasing this (and therefore merging to `develop`) until after we do a point release sometime this week.


larrysalibra:
Merging this in since we already shipped the point release.

PR #1129, opened 2017-12-19
Multi-reader storage auth

yknl:
This adds the `app_index` scope to authentication to enable multi-reader storage. If the scope is requested, the browser will calculate and add the app index file URL to the profile under the `apps` key.

image


shea256:
Let's think of something other than "appear in your profile".

This should clearly communicate exactly what the user is authorizing.

More accurately, beyond reading your basic info, the app wants to "store data on your behalf".

From a technical perspective, the app wants to place data in your profile that has the effect of attesting to keys and URLs, but that is a detail that should be obfuscated away.

yknl:
> Let's think of something other than "appear in your profile".

Yes, we definitely need to come up with something better than "appear in your profile". 

But I feel like "store data on your behalf" could apply to regular (single reader) storage permissions as well. Maybe something like "store data in your profile". 

The other question is do we want to display the permission requests for single reader storage as well.

shea256:
@yknl Would you say the app is storing data in your profile? Isn't it more accurate to say the app is storing a URL and a public key in your profile (yes technically data/bytes but not DATA)? That would lead me to say that the effect of storing the URL is allowing the user to attest to where its storage container is located. And that the effect of storing the public key is allowing the user to delegate to a signing key. The most important part here is delegation and not storage. So we want to convey that something is consequential in terms of authority. "on your behalf" conveys this. Storing data in my profile sounds innocuous. I see your point that "store data on your behalf could apply to single user storage as well. But as a user, I don't really care about that. At the end of the day, single user storage should be a no brainer and shouldn't even be asked about. Meanwhile, signing delegation and location delegation are what we really want to ask about. and in effect this means that an app will be able to act on behalf of you when it comes to manipulating data.

yknl:
What about "Store public data on your behalf"? As a user that has no idea how storage works under the hood, the key difference between single-reader and multi-reader storage is the fact that other people can read "multi-reader" data.

shea256:
@yknl OK "Store public data on your behalf" should be fine, although I'm still a bit confused by the distinction. What constitutes public data? With this lightweight version of multiplayer storage will data be able to be encrypted and thus private?

I also do believe that we shouldn't even have a permission for single-player storage. It's a no-brainer that every app automatically has access to local storage and you could make this statement extend to encrypted data synced to the cloud.

yknl:
My thinking is that when the average users sees "store data on your behalf", they'll click approve without thinking about it. Since most if not all apps will store data. The true implications of multi-reader storage is that data stored by the app **can** be public.

And agreed on second point, there's no need to have permissions for single-reader storage.

larrysalibra:
> My thinking is that when the average users sees "store data on your behalf", they'll click approve without thinking about it.

I agree with @yknl 

I'd say it's more accurate and clearer to say something about how the app is asking to publicize your use of the app on your profile.

If i'm using a blockstack app that is "like Tinder for something really embarrassing" - I'd think twice about letting an app publicize my use of it. If it said "store data on your behalf" - I'd approve that because all apps store data on my behalf - that's what apps do!

> Isn't it more accurate to say the app is storing a URL and a public key in your profile

Lightweight multi-reader storage doesn't do anything with public keys or delegated signing keys. This is something we can address in a future version.

larrysalibra:
For context, this is how the authentication prompt looks with the most recent changes: 
screen shot 2018-01-05 at 6 19 02 pm

larrysalibra:
I do think we can improve the authentication dialog copy a bit. We can always do this on a later iteration, but how about something like:

"Let people look up information you share with this app."

For reference, here's a dialog from icloud where apple keeps a list of apps you've given permission to find you by email (sidenote: i've never actually seen this icloud functionality used it practice).

screen shot 2018-01-05 at 5 47 17 pm

Issue #923, opened 2017-10-16
Add a tooltip to explain what default storage is


larrysalibra:
Problem: People are sometimes confused by what default storage means. They wonder where the data is being stored and are confused by the fact that we earlier told them that they are in control of their storage. If we're storing their data on a server somewhere, we need to explain to them why they're in control. It's because their data is signed and encrypted and eventually will be replicable across multiple cloud storage hubs. Which means data tampering and decryption and censorship are all invalid attack vectors.

Solution: Put a tooltip next to the “use default storage” button that explains what “default storage” is. Explain in this tooltip that their computer encrypts and signs all their data before backing it up to a server in the cloud. For simplicity, we connect them with a reliable server to provide the backups and online access. In the future, more will be available and they will be able to choose from multiple vendors across multiple geographies. Something like this.

wbobeirne:
Closing as stale, the UI has changed since this was made.

PR #1130, opened 2017-12-19
Issue #923: Explain default storage during initial setup

PoliteSM:
Just copy changes to hopefully address #923 

After initial setup, while browsing the storage settings, I realized I didn't have a very good understanding of what 'default storage' meant, or what my storage options are. A quick google search brought up issue #923, so in researching the ticket thought I'd make this pull request.

I changed the link copy of the first storage screen to read 'Alternative storage preview' to better set user expectations that alternative storage is not yet available. The second storage screen title is 'Alternative Storage Preview', and includes a description that attempts to reassure and explain that the default storage system, Gaia, is both private and configurable.


PoliteSM:
meant to open this against the develop branch.

Issue #1123, opened 2017-12-18
As a user, I would like to know how name registration works when I go to register a name


larrysalibra:
Discussion around wording and action items: https://github.com/blockstack/blockstack-browser/issues/1090

@yknl will do acceptance testing

larrysalibra:
https://github.com/blockstack/blockstack-browser/issues/1090

shea256:
Copy to be used as a jumping off point:

>You're about to submit a registration request for `larry.id`. Hope nobody beats you to it!

wbobeirne:
Closing as stale.

Issue #1090, opened 2017-11-21
Add warning/explainer with checkbox during name purchase


larrysalibra:
We need to make it clear to users that they are burning bitcoin and racing other users to buy names. They're not buying names from Blockstack PBC or any other entity (this is decentralized after all!)

Let's do this by adding this language with a checkbox to the name registration process:

Proposed language:
```
I understand by registering a name:
* I am racing other users to register this name by burning bitcoin
* If someone registers the name before me, the bitcoin my computer has burnt in this process will still be gone.  
* No bitcoin from this process goes to Blockstack PBC or any other entity.
```

Feel free to improve on this language.

Any pull request for this issue should have corresponding tests.

larrysalibra:
@jackzampolin we have another issue about sending both of those transactions at the same time: https://github.com/blockstack/blockstack-browser/issues/1091

kantai:
My feelings on this are that we should separate our treatment of this into solving two different problems

1. Informing users of the risks of registration (it's a race, and someone can win, and the loser has still burnt bitcoin) and the process (_your_ computer is doing 2 bitcoin transactions with addresses _you_ control, the burnt bitcoin doesn't go to us)

2. Usage agreement - This is a usage policy that seems to be tied to liability for use of the software. this is a legal decision (that probably interacts with software licensing), that I think should probably be consulted with lawyers before we add a lot of legal text to user's interaction with the product.

larrysalibra:
+1 on @kantai 's analysis. particularly consulting lawyers.

jackzampolin:
I would like to +1 this as well. Both of these things need to be done. It will give users more understanding of how the system works (a good thing) and also will give us a firm policy on these type of issues which makes answering queries about it very simple.

larrysalibra:
I updated the issue with some proposed language.

vsund:
Do we still have preorder transactions to circumvent hijacking claim transactions? The message sounds like other people can also claim your name while registering it.

I didn't look over the new claiming process (with only one transaction), but we still have some technique in place to make sure nobody can listen to transactions and also claim these names, right?

Not really sure whether we should expose all this technical stuff to consumers, but wanted to bring this up though. Should we say that nobody can claim your name when you first claimed it?

larrysalibra:
> Do we still have preorder transactions to circumvent hijacking claim transactions? 
Yes we do.

> The message sounds like other people can also claim your name while registering it.
[...]

PR #1144, opened 2018-01-03
Feature/name registration warning

larrysalibra:
This pull request addresses #1090 & #1123.

It displays a warning screen to the user with information about what happens during name registration.
34298135-1da1696e-e757-11e7-9de1-b109d0c71ae4
34298136-20533444-e757-11e7-8d0f-307f41e83a2d

This should have UI component tests but we haven't yet set up our framework for that.

@yknl can you take a look at this when you have a chance?


shea256:
@larrysalibra Just commented in #1090 about the copy.

yknl:
I agree with @shea256 's comments in #1090. "Hope nobody beats you to it!" seems a little too harsh and lacks context as the title of that modal. Perhaps we should say that name registration requests cannot be guaranteed. There is a small chance that someone may beat you to it.

shea256:
@yknl Yeah and in general I think this copy needs quite a bit of work. We also need continuity of this copy across the modals. Having someone who is just thinking about copy here would be extremely helpful.

yknl:
Proposed copy: @shea256 
registration-disclaimer

shea256:
Very nice work Ken, this is pretty awesome.

A couple of things I was wondering:

1. Should we first mention that registrations are a race or is that clear enough already?
2. Will the user know what "Blockstack PBC" is or should we say no company gets the fee?

If that's the case then it could look something like this:

---

### Name Registration Disclaimer

You're about to create a registration request for myblockstack.id

Please confirm you understand the following:

- Registrations are a race & there's a small chance someone else will win
- Registration requests have a fee regardless of the outcome
- Fees are destroyed on the network and not sent to any company

---

Let me know your thoughts.

yknl:
I think yours is an improvement. I'll make those changes and merge this PR.

yknl:
image

shea256:
Awesome!

larrysalibra:
Great work on this @yknl ! Thanks!

PR #1145, opened 2018-01-04
Add Graphite app to the browser home page.

jehunter5811:


jehunter5811:
@yknl Sorry about that! Made the change.

yknl:
Thanks!

jehunter5811:
@yknl any idea why the icon would be broken? Seems like I had both sizes included in the commit and titled properly...

yknl:
@jehunter5811 are you experiencing the same thing as @dantrevino on v0.22.0? What operating system are you using?

jehunter5811:
@yknl I just downloaded it and it works for me on Mac. Looks great! Thanks!

kantai:
I'm also experiencing this issue --

it looks like the file `app-icon-graphite-256X256.png` should be `app-icon-graphite-256x256.png`

https://github.com/blockstack/blockstack-browser/blob/master/app/images/app-icon-graphite-256X256.png

Maybe the difference on OS's comes down to macOS case-insensitivity ?

jehunter5811:
Good catch. I just tried to commit a new branch and open a pull request, but it didn't look like anything was committed. All I changed was the "X" to "x" so maybe git didn't recognize that as a change?

kantai:
I opened a PR for the name change @jehunter5811 -- we should be able to do another point release to get this working on other OSes.

Issue #1142, opened 2017-12-28
Cannot Log Into Any App On browser v0.21.6


jehunter5811:
I can't tell is this something local to me or impacting everyone, but after downloading the pre-release browser v0.21.6, I cannot log into any Blockstack apps. I am using Chrome and I get the following console error: 

```
Uncaught (in promise) TypeError: Cannot read property 'split' of undefined
    at isLaterVersion (main.js:50534)
    at AuthModal.componentWillReceiveProps (main.js:3889)
    at ReactCompositeComponentWrapper.updateComponent (main.js:153895)
    at ReactCompositeComponentWrapper.receiveComponent (main.js:153828)
    at Object.receiveComponent (main.js:159835)
    at ReactCompositeComponentWrapper._updateRenderedComponent (main.js:154035)
    at ReactCompositeComponentWrapper._performComponentUpdate (main.js:154005)
    at ReactCompositeComponentWrapper.updateComponent (main.js:153926)
    at ReactCompositeComponentWrapper.performUpdateIfNecessary (main.js:153842)
    at Object.performUpdateIfNecessary (main.js:159867)
```
Here are two screenshots of the full log: 

<img width="1363" alt="screen shot 2017-12-28 at 10 19 37 am" src="https://user-images.githubusercontent.com/10519834/34416354-b4009ed8-ebb8-11e7-9205-1e5c09f4f78e.png">
<img width="1366" alt="screen shot 2017-12-28 at 10 19 50 am" src="https://user-images.githubusercontent.com/10519834/34416355-b5fea784-ebb8-11e7-94f5-e0863d42603e.png">

pranavj1001:
even I'm facing the same problem. I fixed it by downgrading to v0.21.5

dantrevino:
Same issue.  I cannot log into apps with 0.21.6.  Using Linux.

GET http://localhost:6270/v1/names/ net::ERR_CONNECTION_REFUSED
(anonymous)
proxyFetchForSomeHosts @ main.js:18327
login @ main.js:3941
invokeGuardedCallback @ main.js:156628
executeDispatch @ main.js:150730
executeDispatchesInOrder @ main.js:150750
executeDispatchesAndRelease @ main.js:150165
executeDispatchesAndReleaseTopLevel @ main.js:150176
forEachAccumulated @ main.js:161919
processEventQueue @ main.js:150376
runEventQueueInBatch @ main.js:156695
handleTopLevel @ main.js:156705
[...]

PR #1164, opened 2018-01-12
assume version zero if auth request has no version

larrysalibra:
This pull request addresses issue #1142 

It checks for the existence of a version field in the authentication request and sets it to `'0'` for the comparison if it doesn't exist. Older versions of blockstack.js didn't have `version` fields.


yknl:
Thanks for getting this one @larrysalibra, tested it out and fixed the issue for me.

PR #1165, opened 2018-01-12
Fix graphite icon on case-sensititve OSes

kantai:
Rename the graphite icon so that case-sensitive OSes can find the file

Addresses issue discussed in #1145


yknl:
Thanks @kantai

Issue #1154, opened 2018-01-09
Writing Profile to Different Locations


kantai:
A subtle result of issue #1112 is that a given identity's profile URL will be written to different locations when the bug is fixed. For example, let's say I have 3 identities each with a profile:

Before bugfix:

```
ID A: Address = m/888'/0'/0'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/0/profile.json
ID B: Address = m/888'/0'/1'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/1/profile.json
ID C: Address = m/888'/0'/3'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/2/profile.json
```

After bugfix:
```
ID A: Address = m/888'/0'/0'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/0/profile.json
ID B: Address = m/888'/0'/1'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/1/profile.json
ID D: Address = m/888'/0'/2'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/2/profile.json
ID C: Address = m/888'/0'/3'
         URL = https://gaia.blockstack.org/hub/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/3/profile.json
```

There's _two_ issues above:
1. ID C's profile URL changed. If there's a name associated with the ID, that means that we'll need a zonefile update. Even if there's not, the user will "lose" their profile data.
2. ID D's profile URL is the same as ID C's old one! So even if we solve the previous problem by allowing ID C to write to the old location, we have an issue where two profiles want to be at the same location.

_Proposed solution_
1. Browser should, in the case of profiles which _don't_ have names and zonefiles, always write to the "correct URL"
2. In the case of profiles which _do_ have names and zonefiles, _try_ to write to the URL in the zonefile. If it cannot, raise an error.
[...]

Issue #1112, opened 2017-12-11
Addresses not generated consistently by browser


larrysalibra:

jackzampolin:
<img width="737" alt="screenshot 2017-12-11 12 21 46" src="https://user-images.githubusercontent.com/7452680/33853273-3b4acac0-de72-11e7-8bf9-3a727451d8ac.png">
<img width="734" alt="screenshot 2017-12-11 12 21 38" src="https://user-images.githubusercontent.com/7452680/33853275-3b686152-de72-11e7-9931-7a543f85edd3.png">

locomo:
i'm getting the same things .. i reset blockstack browser and i expected to see the 3 IDs that I had previously generated but initially only the first address (ID) that I had originally created appeared - but the rest did not reappear on their own 

i clicked "Create New ID" - my 2nd ID appears with the correct address and profile info

click "Create New ID" again .. a new address appears and the profile info for the 3rd ID appears (avatar, name, description) but the username doesn't not appear and it still reads "Add username" and the address is incorrect

if i continue to click "Create New ID" additional addresses are generated but my 3rd {name}.id never appears

Esharifov:
I had the similar issue and currently only have 2 of my 5 ids displaying? 

Would it be better if the user isn't just dependent on UI, and can double verify the ownership of the ids from some other source or claim them by their passphrase or some other solution?

When we purchase the ids, what is it linked to as a primary identifier, is it our account, which we can gain access to with passphrase?

larrysalibra:
> Would it be better if the user isn't just dependent on UI, and can double verify the ownership of the ids from some other source or claim them by their passphrase or some other solution?


You can verify the ids are owned by your identity address by using the blockstack explorer: https://explorer.blockstack.org or spinning up your own full core node.

Esharifov:
Larry I followed same steps as @locomo did, and was able to retrieve all 4 non-default ids.

However when looking into https://explorer.blockstack.org, I noticed that only my default id is populating under my identity, the rest 4 are displaying solo owner and public addresses associated with them.

I guess my question is that, in case BlockStack has issues and doesn't display my info even in the explorer, can I claim these addresses from Bitcoin protocol by entering my passphrase?

jackzampolin:
@Esharifov All of the identity keys for all of those names are derived from your keychain phrase. You can see the details here: https://github.com/blockstack/blockstack-browser/issues/330

jackzampolin:
[...]

PR #1173, opened 2018-01-17
Perform smart profile reads and writes, and fix the address generation bug

kantai:
This addresses #1112 #1154 

Mostly, it performs reads _and_ writes of profiles more intelligently -- as described in #1154.
Reads check any of the valid locations (or the zonefile pointer), using the identity address to validate when the correct location is found. Writes go to either the default location, or wherever the zonefile points, and raises an error if the zonefile location cannot be written to.

In order to test this functionality, get your recovery-phrase ready. Wipe-restore your browser -- profiles with zonefile entries (which are writable) should behave normally. Profiles written to old locations _without_ names should be _read_ correctly, and then the writes should go to the new `/hub/<identity-address/profile.json` location.

This PR _does_ not change the behavior of the redux state, just the way that the state is generated, which I'm pretty sure means that we don't need to perform the state-change upgrade. _However_, if users wish to get at their missing addresses, they will need to reset/restore. So I'm not sure of the best strategy there.


yknl:
Hey @kantai, I ran into the following issue while testing this out:

After saving my index 2 profile, all subsequent addresses seem to be picking up the profile.json of the last address that has a profile.json. 

If I add a profile to my index 3 address, then all subsequent addresses will pick up the profile from index 3.

yknl:
On further testing, it seems to be only happening for adjacent indexes.

kantai:
Thanks for testing @yknl! I'll look into this.

kantai:
I believe this bug came up because I hadn't pushed the merge commit b6d6b790271b8f980921eaeb9e7a5aef87fd57de of `hotfix/address-index-bug` --- the profile read path assumes that no-zonefile profiles for an identity with address `X` of (browser) index `i` come from one of the following locations, given that the 0th address is `A`:

* `{hub}/{X}/profile.json`
* If `i in [0,1]`:  `{hub}/{A}/{i}/profile.json`
* If `i` is odd and `>1`: `{hub}/{A}/{1 + (i divided by 2)}/profile.json`

That logic only works if the browser was reset -- otherwise the location `{hub}/{A}/{1 + (i divided by 2)}/profile.json` should _also_ be a possible location -- we can add that path as well to the possible set if we don't want users to have to reset their browser.

But anyways, the code _should_ have only succeeded in reading a profile for a given identity if the `profile.json` was validly signed by that identity -- do you have any screenshots of the behavior so I can look more into what was going on there?

yknl:
@kantai It looks like this is a separate bug that also exists in develop.

yknl:
@kantai I submitted a fix for the above mentioned bug that I found during testing. If you pull it from develop we can test it out with the changes in this PR.

kantai:
Excellent, thanks @yknl!

Just tested this out and it works for me

larrysalibra:
Taking a look at this now!

larrysalibra:
I'm having trouble with one of the tests timing out:

screen shot 2018-01-24 at 9 59 22 pm

I suspect it's because the test doesn't have mocks for the profile reads so they are hitting the network.

larrysalibra:
@kantai added a commit to mock those requests and configured `nock` to throw an error if there's unmocked network access.

kantai:
Awesome! Thanks for the reviews @yknl and @larrysalibra (and thanks for fixing the mock URLs I was missing!)

Merging to develop now.

Issue #1186, opened 2018-01-26
Remove legacy dropbox and s3 related code


larrysalibra:
As part of storage technical debt clean up, we want to remove legacy dropbox and s3 related code #1184 .

larrysalibra:
This was merged.

Issue #1185, opened 2018-01-26
Chrome doesn't load browser when a chrome update is pending


larrysalibra:
I ran into this issue today and wanted to document in case someone else runs into it:

When a chrome update on macOS has downloaded and is pending install, opening the Blockstack Browser from the mac app gives you a blank chrome window with no page visible. 

It's not just a blockstack problem, any app calling the default browser open function results in a blank page being opened in chrome.

Making this as wontfix since it isn't a problem on blockstack's side.

Moving to a native solution would solve this.

bionicDan:
I ran into this issue myself as well, and only determined the cause after finding the issue documented here.

Since the issue itself is out of our control, suggest preventing the failure from happening _silently_ from the user's perspective. Possible solutions (feasibility not yet determined) indicating the issue in the browser window itself, indicating the issue in the Mac's Blockstack menu bar item or documenting the potential for this behavior in the installation instructions.

wbobeirne:
Now that it's a native app, this issue shouldn't persist.

Issue #1184, opened 2018-01-26
Storage technical debt clean up


larrysalibra:
- [ ] #1186 Remove legacy dropbox & s3 related code 
- [ ] #1187 Simplify gaiahub connection to not require redirects

wbobeirne:
Looks like this was addressed in #1188.

PR #1188, opened 2018-01-26
Remove legacy dropbox and s3 code

larrysalibra:
This pull request addresses #1186 which is part of the storage technical debt cleanup #1184 in the browser. 

It removes dropbox and s3-related code and refactors tests.

It does not change the redirect code that is used to connect gaiahub (#1185).

To test this, you'll want to make sure that on-boarding works - specifically connecting storage and that operations that use storage such as uploading photos and profiles and apps work.


larrysalibra:
Thanks for taking a look at this @yknl !

Issue #1195, opened 2018-01-31
Remove auth support for apps that require core storage


larrysalibra:
We have legacy support for apps that used core storage. This was where apps had to write via a localhost core endpoint.

This legacy support only worked on macOS. We've switched to a direct to gaiahub model which doesn't require core as an intermediary.

We should remove support from the authentication process for apps that use core storage.

https://github.com/blockstack/blockstack-browser/blob/9c776fa9eea92b3b76f538ebab597a84e2e93c0e/app/js/auth/components/AuthModal.js#L380

Issue #1194, opened 2018-01-31
`apps` key not populated properly


larrysalibra:
Moved from https://github.com/blockstack/blockstack.js/issues/353 reported by @bodymindarts

Having issues with multi-player read. The apps key isn’t being populated properly for @kp8383

blockstack.lookupProfile("kyle.id").then(profile => console.log(profile.apps));
kyle has logged into an app with the publish_data permission and has written a file but its not accessible because of ^^^.

Also the error message when this happens isn't very descriptive:

> blockstack.getFile("public.json",{username: "kyle.id",app:"http://localhost:3000"}).then(text => console.log(text));
> Promise {<pending>}
> Uncaught (in promise) null
Just having null as the error makes finding the root cause tricky.

larrysalibra:
This was fixed in #1196 and shipped in v0.23.0.

PR #1196, opened 2018-01-31
Fix apps key doesn't populate issue #1194

larrysalibra:
This pull requests fixes 3 problems that were causing the apps keys not to populate to the user's profile.json file when an app requested multi-player storage.

1. the signed profile token data was generated before the new apps array was generated
2. the uploadProfile function was being passed the identity index instead of the identity object
3. the uploadProfile function was being passed the identity address instead of the signing key

While looking into this issue, I also discovered that the `setCoreStorageConfig` function has not been updated to use the new signature of `updateProfile` that changed during the fixes for address generation. We're not going to be supporting core storage going forward so suggest we remove it: #1195.


larrysalibra:
Thanks @kantai and @yknl for your assistance on this!

PR #1198, opened 2018-02-01
Fix Windows and Linux instructions on checklist

kantai:


cwackerfuss:
These changes look a-okay to me! What's the status of this ticket? Since I don't have experience with Windows and Linux setup I'll let @larrysalibra & @yknl decide on approval.

Issue #1193, opened 2018-01-30
Re-enable subdomain support in the browser


larrysalibra:
We previously added subdomain support in #664 and then disabled it in #686 when we didn't have bandwidth to ship production subdomains.

We'll need to re-enable subdomain support.

PR #1204, opened 2018-02-06
#1193 Enable subdomain user registration

cwackerfuss:
Completes #1193 

subdomain-registration

To test code:
1. Run `npm install` to install Enzyme and its dependencies for React component testing
2. Verify tests pass using `npm run test`

To test UI:
1. Navigate to `http://localhost:3000/profiles/i/add-username/0/search` in development mode
2. Enter a username (if one does not exist)
3. See that the app searches through the subdomain records and shows that you can register a subdomain with your username.


larrysalibra:
@cwackerfuss I'm having a bit of trouble getting my regtest environment and it's getting late so I'll have to take a look at this tomorrow.

Did you try this with the regtest environment and register a subdomain to make sure everything works normally with a blockstack id that has a subdomain?

cwackerfuss:
@larrysalibra I did finally get Subdomain Registrar working from regtest with a lot of help from the team. Here is what got it working for me: https://github.com/blockstack/blockstack-core/pull/737

I've confirmed that subdomains get registered correctly and appear on the user's account.

screen shot 2018-02-08 at 3 11 24 pm

larrysalibra:
@cwackerfuss was your subdomain registration ever processed? (ie. did the registration in process clock icon disappear?)

screen shot 2018-02-10 at 3 03 17 am

Mine never showed up on the regtest core node.

I do see that the registrar is receiving it though, so I don't think this is a problem with browser code.

screen shot 2018-02-10 at 3 04 51 am

If registration didn't complete for you either, let's open an issue about this on core.
--

Two copy issues:

Users don't need to keep their computers on for subdomain registration, so we can show different copy. All we need to communicate is that their name should be ready for use in about an hour.

screen shot 2018-02-10 at 2 58 59 am

Calling these "subdomain" usernames "subdomains" is confusing. See discussion: https://github.com/blockstack/blockstack-browser/issues/1211

What about changing the below to something like "**orange.foo.id** is a free username."?

screen shot 2018-02-10 at 3 03 51 am

cwackerfuss:
This is ready for you again @larrysalibra !

larrysalibra:
Thanks @cwackerfuss - taking a look now!

larrysalibra:
This looks great @cwackerfuss ! 

screen shot 2018-02-14 at 7 31 42 pm

Ready to merge!!!

Issue #1207, opened 2018-02-07
Backfill JS Linting


cwackerfuss:
We added the linting task to the js app late in the process, so there are files that are [currently being ignored](https://github.com/blockstack/blockstack-browser/blob/master/.eslintignore) that we have the opportunity to go back and lint.

cwackerfuss:
@larrysalibra and @yknl --

I'm going to take a breather on going through the Linting fixes so that
A) we dont have too many open tickets
B) there's less chance of merge conflicts

Please take a look and review when you can, I know they're technically lower priority tickets but the fact that they touch so much code makes me want to get them reviewed and merged as soon as possible (because conflicts).

larrysalibra:
@cwackerfuss sounds good.

also making a note to add linting coverage to our test suite as well

wbobeirne:
Closing as stale.

PR #1208, opened 2018-02-08
JS Linting Account Components

cwackerfuss:
#1207 
Satisfies the backfilling of linting fixes for Accounts components

This PR looks huge -- but most of it is indentation changes.

NOTE:
I disabled the `react/sort-comp` rule because I feel most developers do not have a problem following this rule by nature, and it can be too strict in its order requirements around click handlers and various class method types
[Read about react/sort-comp here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md)
We can have a discussion here around whether or not we want to enforce this rule.

TO TEST:
Please pull down and navigate through the accounts section of the app to test that features still work.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -15,13 +15,6 @@ tests/utils/a*
 tests/utils/e*
 tests/utils/n*
 
-# app/js/account/A*
-# app/js/account/B*
-# app/js/account/Ch*
-# app/js/account/Cr*
-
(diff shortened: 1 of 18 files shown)


yknl:
Yea I think it's a little too strict as well. I'm onboard with disabling it.

PR #1210, opened 2018-02-08
JS Linting Utils

cwackerfuss:
#1207
Satisfies the backfilling of linting fixes for Utils files

TO TEST:
You can regression test the application. Since this touches files in app/js/utils, the effects are wide-reaching, but if you take a look at the diff, you'll see there aren't any really smelly updates.

PS I manually tested this change from:
```if (response.status == 404) {```
to
```if (response.status === 404) {```

I wasn't sure if `response.status` was of type `number`. You get two different results from numbers vs. strings with this logic. Turns out, `response.status` IS of type `number`, so we're fine.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -26,9 +26,6 @@ app/js/components/*
 !app/js/components/Navbar*
 
 app/js/store/*
-app/js/utils/*
-!app/js/utils/account-utils*
-!app/js/utils/api-utils*
 
 app/js/profiles/store/*
 app/js/profiles/tabs/*
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,9 @@
   "plugins": [
     "react"
   ],
+  "globals": {
+    "proxyFetch": true
+  },
   "ecmaFeatures": {
     "jsx": true,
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

File: .eslintignore [hunk 1]
```diff
@@ -26,7 +26,7 @@ app/js/components/*
 !app/js/components/Navbar*
 
 app/js/store/*
-app/js/utils/*
+# app/js/utils/*
```
larrysalibra: we can remove the 3 lines that start with `app/js/utils` now that utils are lint free.

cwackerfuss:
Good callout @larrysalibra -- updated the branch to remove the eslint ignore lines

larrysalibra:
@cwackerfuss in the future, you can just push an additional commit to address review comments. please don't rebase published commits. 

From https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing

screen shot 2018-02-09 at 11 43 45 pm

When I tried to pull in your updated, I ended up with conflicts between your previous commit that no longer exists in this branch and new rebased commit.

screen shot 2018-02-09 at 11 44 15 pm

PR #1220, opened 2018-02-20
Remove email scope whitelist

larrysalibra:
We previously restricted access to the `email` scope to apps in our whitelist. 

There is strong interest in the developer community to have access to this functionality so that they can pre-populate email fields with user emails.

This pull request removes the whitelist.

I'm opening it to start a discussion around whether we want to merge this quick and dirty change so that developers can get started or if we want to do additional work: how this is communicated to users, giving users option to opt out, etc.

### DIFF
--- a/app/js/auth/components/AuthModal.js
+++ b/app/js/auth/components/AuthModal.js
@@ -24,18 +24,6 @@ import { signProfileForUpload } from '../../utils'
 
 const logger = log4js.getLogger('auth/components/AuthModal.js')
 
-const APP_EMAIL_SCOPE_WHITELIST = [
-  'https://staging.blockstack.clients.barefootcoders.com',
-  'https://blockstack.com',
-  'http://localhost:3000',
-  'http://localhost:3001',
-  'http://localhost:3002',
-  'http://localhost:3003',
-  'http://localhost:4000',
-  'http://localhost:5000',
-  'http://localhost:8080'
-]
-
 function mapStateToProps(state) {
   return {
     localIdentities: state.profiles.identity.localIdentities,
@@ -111,15 +99,8 @@ class AuthModal extends Component {
     const authRequest = getAuthRequestFromURL()
     const decodedToken = decodeToken(authRequest)
 
-    const appDomain = decodedToken.payload.domain_name
     const scopes = decodedToken.payload.scopes
-    let invalidScopes = false
-
-    if (scopes.includes('email')
-    && !APP_EMAIL_SCOPE_WHITELIST.includes(appDomain)) {
-      logger.error(`componentWillMount: ${appDomain} not in 'email' scope whitelist`)
-      invalidScopes = true
-    }
+    const invalidScopes = false
 
     if (scopes.includes('email')) {
       this.setState({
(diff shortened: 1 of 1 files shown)


larrysalibra:
Additional background info:

There were a few reasons we had this whitelist in the first place:

* the authentication dialog language isn't very clear that their email is being sent to apps that request this (it adds the word "email" as something being sent)
* there's not a way for the user to opt out or use a different email
* there's not a way for user to change their email stored in the browser
* Not every user has an email in the browser (older users could skip providing an email

Would like some feedback on these changes from a product perspective before we go forward with them.  See https://github.com/blockstack/blockstack.js/issues/377 as well.

tagging @yknl @shea256 @pstan26 @kantai

larrysalibra:
@cwackerfuss would love if you could take a look at this

app/js/auth/components/AuthModal.js
```diff
@@ -111,15 +99,8 @@ class AuthModal extends Component {
...
-    const appDomain = decodedToken.payload.domain_name
     const scopes = decodedToken.payload.scopes
-    let invalidScopes = false
-
-    if (scopes.includes('email')
-    && !APP_EMAIL_SCOPE_WHITELIST.includes(appDomain)) {
-      logger.error(`componentWillMount: ${appDomain} not in 'email' scope whitelist`)
-      invalidScopes = true
-    }
+    const invalidScopes = false
```
cwackerfuss: We can safely remove `invalidScopes` here and remove it from the `setState` in `componentWillMount`. `invalidScopes` is already being set to `false` in the constructor

app/js/auth/components/AuthModal.js
```diff
@@ -111,15 +99,8 @@ class AuthModal extends Component {
     const authRequest = getAuthRequestFromURL()
```
cwackerfuss: Cleaned this up a bit, tell me what you think:

```
  componentWillMount() {
    const authRequest = getAuthRequestFromURL()
    const decodedToken = decodeToken(authRequest)
    const { scopes } = decodedToken.payload

    this.setState({
      authRequest,
      decodedToken,
      scopes: {
        ...this.state.scopes,
        email: scopes.includes('email'),
        publishData: scopes.includes('publish_data')
      }
    })

    this.props.verifyAuthRequestAndLoadManifest(authRequest)
  }
```

cwackerfuss:
Aside from the PR feedback I left, just wanted to suggest we make it more transparent in the new onboarding flow that your email address will be used for more than just recovery email. We should also definitely support modifying email addresses after an id has been registered if we're going to be using them in this manner.

app/js/auth/components/AuthModal.js
larrysalibra: This looks great! Can you make the change?

larrysalibra:
>Aside from the PR feedback I left, just wanted to suggest we make it more transparent in the new onboarding flow that your email address will be used for more than just recovery email. We should also definitely support modifying email addresses after an id has been registered if we're going to be using them in this manner.

Yes! I'll defer to you and @jeffdomke on how to do this - seems like the new on-boarding flow is the place to take care of this.

@cwackerfuss Changing ownership of this pull request to you - feel free to merge it once you make the change you suggested above! Thank you!!

PR #1221, opened 2018-02-21
Patches for client-side transaction generation

kantai:
This does three things:

1. Sets the 6270 port in the library explicitly (we can change the library to default to 6270 if we want, but this makes the browser changes work now, and it won't break if we change the default)
2. Adds the register transaction's spends and unspents to the UTXO set, which will keep the register's inputs from getting spent before it has been broadcasted.
3. Performs safety checks for register: name is valid and address can receive a name


yknl:
thanks @kantai! merging this in.

PR #1222, opened 2018-02-22
Swap Bootstrap for Blockstack in issue template

nsipplswezey:


larrysalibra:
Awesome catch! Thank you!

PR #1226, opened 2018-02-25
handle other python2.7 paths, upgrades venv pip

nsipplswezey:
PR handles two small issues I ran into while getting set up:
- having python2.7 at a different path than `/usr/bin/python2.7`
- python2.7 using a version of pip that didn't have the `--only-binary` flag necessary `pip install cryptography --only-binary cryptography`
Hope the generic patch name is ok, since this is small enough to do through github UI. I can also make the PR to the `develop` branch if that's preferred.

### DIFF
--- a/native/macos/build_blockstack_virtualenv.sh
+++ b/native/macos/build_blockstack_virtualenv.sh
@@ -22,7 +22,10 @@ rm -Rfv blockstack-venv
 
 echo "Creating a new virtualenv..."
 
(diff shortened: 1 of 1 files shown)


nsipplswezey:
After reading https://github.com/blockstack/blockstack/tree/feature/engineering-handbook/engineering it seems the `develop` branch is the correct target for this change.

larrysalibra:
Thanks for this improvement and for reading the engineering handbook :-D

Issue #1228, opened 2018-02-26
Wrong profile URL passed to apps when logging in without a username


kantai:
@meteorplus posted this issue over on https://github.com/blockstack/blockstack.js/issues/383, but it's more appropriate here:

>You can see that the profile url return from the aut response is wrong and has an index in it and shoul dbe /profile.json not /0/profile.json
>![blockstack error](https://user-images.githubusercontent.com/12363020/36651387-a8b81990-1aa0-11e8-96e1-5cd670282f06.jpg)
>
>
>and because of this I am fetching the profile file a second time after the userLoadData just to get the full name and the profile picture.
>
>also some this assume the user already has a username while many don't and just have an address
https://github.com/blockstack/blockstack.js/blob/30175c820c3e24fa3223a332304b690d7c1b408f/src/auth/authVerification.js
>
>Thank you :)

I believe the issue is due to the fact that when we changed the profile URLs to use the actual identity address, rather than an index off of the 0th index, we didn't update the auth code in browser.

kantai:
Included in release v0.26.0

PR #1229, opened 2018-02-26
fix the profile URL used for auth logins without a username.

kantai:
This changes the AuthModal to use the same profile lookup path as the profile editing components for identities _without_ a user name. This moves the `fetchProfileLocations()` function to `profile-utils`, adding a URL to its return. The AuthModal now calls this to obtain the default profile location.

Test via:

0. Run app `npm run dev` or `npm run dev-webapp` (if you're lazy like me and don't want to start a blockstack API endpoint up).
1. Setting up an identity address with a name in the profile, but _no_ domain/subdomain associated with it.
2. Log into Blockstack Todos, you should see "Your Name's Todos"
3. Log out of Blockstack Todos
4. Log into Blockstack Todos _with_ an identity with an associated domain/subdomain -- you should see the correct information from that profile as well (this is an ad-hoc regression test)

Those above test steps lead me to believe that we want better auth testing in the blockstack-browser test suites sooner rather than later.

This PR addresses issue #1228

### DIFF
--- a/app/js/auth/components/AuthModal.js
+++ b/app/js/auth/components/AuthModal.js
@@ -12,6 +12,7 @@ import {
 } from 'blockstack'
 import Image from '../../components/Image'
 import { AppsNode } from '../../utils/account-utils'
+import { fetchProfileLocations } from '../../utils/profile-utils'
 import { setCoreStorageConfig } from '../../utils/api-utils'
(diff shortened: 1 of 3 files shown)


kantai:
Poking @yknl for a review on this before the next browser build. This bug prevents user profiles from loading correctly when logging into applications without a username in the current blockstack browser.

PR #1232, opened 2018-02-27
Make dotpodcast user ready app

yknl:


app/js/data/apps.js
```diff
@@ -117,10 +117,10 @@ const apps = {
         small: 'app-icon-dotpodcast-256x256.png',
         large: 'app-icon-dotpodcast-1024x1024.png'
       },
-      website: 'https://dotpodcast.co/',
-      launchLink: 'https://dotpodcast.co/',
+      website: 'http://player.dotpodcast.co',
```
larrysalibra: confirming with the app developer that they don't want https://player.dotpodcast.co (with https)

yknl: good catch

Issue #997, opened 2017-10-30
Open an Ethereum block explorer when you click on an Ethereum address in a profile


shea256:

larrysalibra:
I think @cponeill is taking a stab at this in #996

Issue #996, opened 2017-10-30
Open a Bitcoin block explorer when you click on a Bitcoin address in a profile


shea256:

larrysalibra:
@cponeill how's this coming?

cponeill:
@larrysalibra I started this and then got wrapped up in another project. I can definitely jump back into this and give it a go. I may have some questions about it.

PR #1234, opened 2018-03-04
On the profile, open bitcoin / ethereum explorer when clicking on add…

ramfordt:
Proposal for #996 and #997: opens block explorer when the user clicks on a Bitcoin or Ethereum address on a profile, and profile is not in edit mode


yknl:
Closing as this is out of date

PR #1248, opened 2018-03-14
Hotfix for Proxy Bugs

kantai:
This hotfix patches the blockstack proxies:

1. to default bind to `localhost`, rather than `0.0.0.0` (it also changes the Dockerfile so that it will set the bind to 0.0.0.0).
2. to check for attempted path traversals and return the `index.html` in those cases.

We should consider a point release for these fixes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,6 @@ RUN /src/blockstack-browser/node_modules/.bin/gulp prod
 
 # Setup script to run browser
 RUN echo '#!/bin/bash' >> /src/blockstack-browser/blockstack-browser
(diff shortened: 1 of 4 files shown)


larrysalibra:
Sorry about that - just realized that you added this as a 4th parameter with a default. Making a build of the mac app to verify everything works.

larrysalibra:
Thanks for taking a look at the `nexe` bug that was preventing blockstackProxy from building on macOS. @yknl

larrysalibra:
> We should consider a point release for these fixes.

I agree.

kantai:
Thanks for looking at this @yknl and @larrysalibra 
I'll bump the `package.json` version, and then will merge

Issue #1249, opened 2018-03-14
As a user, I should be able to view the chat app bounty winners in the Blockstack browser app store


cwackerfuss:
The main question here is where should the chat apps live? In an existing section or in their own new section? Discuss!

cwackerfuss:
Cc @yknl @shea256 @jeffdomke @larrysalibra

yknl:
I vote for a new section

cwackerfuss:
Making a new section unless @shea256 @jeffdomke @pstan26 @larrysalibra @xanbots object

jw2013:
@cwackerfuss shall we close this issue as https://github.com/blockstack/blockstack-browser/pull/1250 is merged? :)

wbobeirne:
We shall!

PR #1250, opened 2018-03-15
Add Chat Apps into browser

cwackerfuss:
References #1249

screen shot 2018-03-15 at 12 11 25 pm

1. Added Hermes and Stealthy into the browser under a new section called "Chat Apps"
2. Added `target="_blank" to all app icons so that when user clicks an app, it opens a new tab


jackzampolin:
Was just playing with these apps today! Looks good to me. 👍

yknl:
Looks good to me too!

Issue #1240, opened 2018-03-09
Prevent double-spending of REGISTER transaction once submitted to broadcaster


yknl:
As @kantai mentioned in the client-side transaction generation PR
> Regarding the "name registration success" message, we should add a message that users need to be careful trying to spend out of the payment address while the transactions are in flight -- they risk double-spending the REGISTER transaction. Eventually, we'll want to make sure that the displayed balance / bitcoin transaction generation process can't do that (we'll need to persist those transactions inputs/outputs locally, likely in redux state).

1) We should add warnings in the "name registration success" screen as well as in the wallet screen when a registration is pending. 

2) Persist the name registration transactions inputs/outputs locally and check against that when generating new name registration transactions.

kantai:
So, this already looks like it's created some issues to me -- I'm going to start by changing the transaction broadcaster so that it will _refuse_ to accept a transaction with inputs that will consume another transaction in its queue's inputs.

kantai:
Okay -- added and deployed:

https://github.com/blockstack/transaction-broadcaster/pull/3

However, this won't prevent a user from borking their transactions with a bitcoin spend.

kantai:
This wasn't really resolved -- it just gets rejected by the transaction broadcaster, but it's not exactly a great user experience.

PR #1255, opened 2018-03-21
Adds warnings while registration pending

kantai:
Adds a warning message to transaction sending views when a registration is pending. This partially addresses issue #1240 by warning users that withdrawing bitcoin or trying to submit another registration while their other registration is pending risks interfering with their other registration.

We can remove these warnings once the browser is smart enough to adjust a user's UTXO set or display more information about pending operations.

## Testing

You can test this in mainnet by submitting a registration, and then trying to submit another one, or view the bitcoin spend page.

You can test this in regtest by firing up the docker regtest container:

```
IMAGE=$(docker run -dt -p 6270:6270 -p 16269:16269 -p 18332:18332 -e BLOCKSTACK_TEST_CLIENT_RPC_PORT=6270 -e
 BLOCKSTACK_TEST_CLIENT_BIND=0.0.0.0 -e BLOCKSTACK_TEST_BITCOIND_ALLOWIP=172.17.0.0/16 quay.io/blockstack/integrationtests:mas
ter blockstack-test-scenario --interactive 2 blockstack_integration_tests.scenarios.browser_env)
```

Running the browser:

```
npm run dev-proxy & npm run dev
```

And opening: `http://localhost:3000/#coreAPIPassword=blockstack_integration_test_api_password&regtest=1`

Then you can submit a registration (after loading up your wallet with `core-send`) and check out the warnings.

Screenshots:

warn-register
withdraw


larrysalibra:
taking a look at this now!

kantai:
I agree that we should not allow users to submit these operations, but only if we _really_ know that they will break a registration. Currently, this just iterates to see if there are any 'pending' names (as denoted by the redux state). Unfortunately, that's not always truthful, because a registration that fails will remain pending forever. That's a problem that should also be solved, but will require a lot more intelligence on the client (tracking the sent transactions, seeing if they are in the mempool, how long has it been since the preorder confirmed?)

In the case we deal with now, the browser doesn't know _for sure_ that a registration will break if the user submits a registration or bitcoin spend, hence, just a warning.

larrysalibra:
@kantai that makes sense to me. Changing my review to approve.

kantai:
Thanks @larrysalibra -- merging now.

PR #1256, opened 2018-03-21
Remove dependency on local core endpoint

kantai:
This PR removes `blockstack-browser`'s dependency on the local core endpoint and enables the bitcoin wallet and name registration processes in the Windows and the webapp builds. These are now identical to the normal build, except that `isCoreEndpointDisabled` will _always_ return `true` on those builds-- on the normal build, when using the regtest environment, that returns `false`.

This PR incorporates changes from the `feature/client-side-btc` #1254 (that needs to be approved before this should be), but not #1255.

# Testing

### Regtest Testing

To test in regtest, start the docker container:

```
$ IMAGE=$(docker run -dt -p 6270:6270 -p 16269:16269 -p 18332:18332 -e BLOCKSTACK_TEST_CLIENT_RPC_PORT
=6270 -e BLOCKSTACK_TEST_CLIENT_BIND=0.0.0.0 -e BLOCKSTACK_TEST_BITCOIND_ALLOWIP=172.17.0.0/16 quay.io/blockstack/inte
grationtests:master blockstack-test-scenario --interactive 2 blockstack_integration_tests.scenarios.browser_env)
$ docker logs -f $IMAGE | grep -q 'Test finished'
```

Once that's initialized, you can run the browser proxy with:

```
$ npm i && npm run dev-proxy & npm run dev
```

If you open `localhost:3000/#coreAPIPassword=blockstack_integration_test_api_password&regtest=1`, it will activate regtestmode, and you should be able to register names, send bitcoin transactions, etc. (once you've loaded up the browser's client-side wallet from the `wallet/send-core` page).

### Mainnet testing

Run the browser, with the dev-proxy, and open `localhost:3000`. You _should_ reset your browser's localStorage, because this build changes around the default API endpoints.

Now -- you should be able to log into apps, just as before, though applications that use deprecated storage will no longer work -- try coinkitty:

old-gaia


kantai:
I'm tagging lots of people for reviews, because this is a big-ish change. But we should wait until the other PRs are reviewed, merged, before starting a review on this, I think.

jcnelson:
What feature branch do I need to check out to test this on Linux?  Asking for the public as well as for myself.

kantai:
If you checkout this branch and run it as follows:

```bash
$ cd blockstack-browser
$ git checkout feature/enable-reg-webapp
$ npm i
$ npm run dev-proxy & npm run dev
```

You'll get a browser at localhost:3000

kantai:
Also -- you can test with the docker script in this issue: https://github.com/blockstack/blockstack-browser/issues/1267

kantai:
Running into some issues now with enabling regtest mode, will have to work on that tomorrow.

jcnelson:
I can't seem to get it to run in regtest.  I don't see any indication in the console that the browser has detected that it is in regtest mode, despite having used the URL `localhost:3000/#coreAPIPassword=blockstack_integration_test_api_password&regtest=1` and having gone through the onboarding process in an Incognito window.

app/js/utils/api-utils.js
```diff
@@ -39,8 +41,82 @@ export function getRegTestModeFromURL() {
...
+    const toFindUrl = 'http://localhost:6270'
+    const toSetUrl = DEFAULT_CORE_API_ENDPOINT
+    const coreAPIPassword = DEFAULT_CORE_PHONY_PASSWORD
+    const outApi = findAndSetApiUrls(api, regTestMode, coreAPIPassword, toFindUrl, toSetUrl)
+    console.log(JSON.stringify(outApi))
+    return outApi
+  } else {
+    return api
+  }
+}
```
cwackerfuss: I'm not a huge fan of how `api` is just passed through this function if the version is > 13. Can we split this out into two functions, called something like:
1. `hasLegacyCoreStateVersion` - returns true or false depending on `existingVersion < 13`.
2. `addLegacyCoreEndpoints` - handles all the logic in the `if` block

Then when this is called, use this logic instead:
```
const api = hasLegacyCoreStateVersion()
  ? addLegacyCoreEndpoints(state.settings.api)
  : state.settings.api
```

kantai: Yep! Will change that

kantai:
Thanks for the reviews, merging to `develop` now...

PR #1258, opened 2018-03-24
Fixes empty error message if you enter an invalid password when registering a username

poffdeluxe:
In the `RegistrationSelectView`, when you attempt to register the username but have entered an incorrect password, an alert area appears but it doesn't show any error text. This is because the alert area text is set to `error.message` but the return value from `decryptBitcoinPrivateKey` (and `decryptMasterKeychain`) is a string and not an error object.

error

Was thinking of just fixing this by changing `this.updateAlert('danger', error.message)` to `this.updateAlert('danger', error)` but it seems better practice to reject the original promise returned by `decryptMasterKeychain` with an actual error object. `decryptMasterKeychain` is used in two places in the repo and both places are already capable of handling an error object.

I also added in a super simple test case for `decryptMasterKeychain` to make sure it's still working & returns an error object when an error happens.

blockstack_browser

Anyway, this is my first contribution to Blockstack so lemme know if I need to format something differently. 😀

### DIFF
--- a/app/js/utils/account-utils.js
+++ b/app/js/utils/account-utils.js
@@ -142,7 +142,7 @@ export function decryptMasterKeychain(password, encryptedBackupPhrase) {
       },
       error => {
         logger.error('decryptMasterKeychain: error', error)
-        reject('Incorrect password')
+        reject(new Error('Incorrect password'))
       }
     )
   })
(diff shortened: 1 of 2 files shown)


kantai:
Thanks for this @poffdeluxe. Testing this out now!

kantai:
This works great for me! And awesome job including a unit test. Will merge into develop now.

Issue #1274, opened 2018-04-09
Error when authenticating an app without a username


cwackerfuss:
An app developer is running into an issue where they can't authenticate when an identity doesn't have a username. I believe this is just a bug that happens sometimes, but may have to do with specific configurations.

<img width="989" alt="screen shot 2018-04-09 at 12 43 59 pm" src="https://user-images.githubusercontent.com/4405356/38513388-b9da5c28-3bf3-11e8-829d-4b5a5cc5f453.png">

kantai:
Thanks for creating an issue for this @cwackerfuss -- I think a couple of others have encountered this issue as well. I have a fix for this, will test and then we can hopefully include this in the upcoming release.

aulneau:
+1 I have run into this, too

kantai:
This is fixed in the latest release (0.28)

PR #1275, opened 2018-04-09
Fix authentication with empty profiles

kantai:
This addresses #1274 --

A profile which hasn't been written yet 404s in the `fetchProfileLocations` function ---  leading to a `null` output, which ultimately breaks the auth modal.

This PR addresses that by checking for a null return and returning the default profile URL in that case.

You can test this by running this branch of the browser, and:

1. Test logging in with a brand new profile, which hasn't written it's profile yet (i.e., you haven't saved any changes to the profile)
2. Test logging in with a saved profile (a profile that _has_ already been written)


kantai:
Thanks @yknl!

PR #1287, opened 2018-04-18
New readme instructions

kantai:
We no longer need to install and setup core api to run the browser


kantai:
Thanks, @yknl, merging.

PR #1292, opened 2018-04-18
Eslint Fix: Ex* components

kkomaz:
Eslint Fix: Ex* components


kantai:
This looks great @kkomaz, thanks.

PR #1295, opened 2018-04-18
[ESLint] Linting for tests/account/*

andrerfneves:
## Tasks
- [x] Removed `tests/account/*` from `.eslintignore`
- [x] Added `max-len` rule support for longer strings and comments
  * addresses tend to be rather large strings
- [x] Sorted ESLint rules alphabetically
- [x] Code clean up and linting efforts for all `tests/account/*` files

## Discussion
Wasn't sure how to handle the edge cases for `max-len`. When the title of a an `it` test description is longer, ESLint still triggers an error. I'm locally disabling those for now. Happy to discuss if you'd like to extend `max-len` to 100, or follow another approach.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,7 +7,6 @@ build/*
 gulp/*
 flow-typed/*
 
-tests/account/*
 tests/fixtures/*
 tests/profiles/*
 tests/store/*
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,10 +20,13 @@
     "eol-last": 2,
     "import/no-duplicates": 2,
     "max-len": [
-      "error",
+      2,
+      100,
+      2,
       {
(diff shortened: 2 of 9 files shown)


andrerfneves:
screen shot 2018-04-18 at 7 39 32 pm
I'll push a fix in a little while

kantai:
Hi @andrerfneves --

Thanks for this PR.

For now, we'd prefer to keep the existing linting rules. It's true that long strings like URLs can frequently end up being longer than the max-len rules, but for cases like this, we break the string up into multiple lines. I'm going to patch that and then merge this into develop.

andrerfneves:
@kantai Yeah that's fair. Wasn't sure how the internal team wanted to handle the `max-len` issue. Thanks for checking and merging.

PR #1296, opened 2018-04-18
Eslint Fix: app/js/profiles/Reg*

littlespex:
Fix lint app/js/profiles/Reg*

contribute.blockstack.org
dococch


kantai:
Tested this out, and it works great for me. Thanks for this-- merging now.

PR #1302, opened 2018-04-21
(fix) linting on ActionItem component

15chrjef:
Fixi linting for blockstack browser via contribute.blockstack.org task.

The solution consisted of restructuring the ActionItem component as pure react component.


yknl:
Closing as ActionItem.js is already delinted

PR #1308, opened 2018-04-29
Show private key on keychain backup page, fixes #1261

### LINKED ISSUE #1261: Reveal private keys within the Browser
For running the Subdomain-Registrar (and other purposes) it would be helpful if we could reveal the Identity's private keys from within the Blockstack Browser similar to the 'blockstack wallet' command in the CLI. Besides disclosing the ownerKey and paymentKey it would make sense to reveal the dataKey as well.
Note that there is a [workaround](https://forum.blockstack.org/t/obtaining-identity-keys-from-the-browser/4663/4?u=blockpac) available at present.

hstove:
On the keychain backup page, this shows your private key in WIF format. This is in response to #1261.

screenshot


yknl:
Thanks @hstove! Merging this one.

PR #1346, opened 2018-05-10
Feature/short encryption

jcnelson:
This pull-request changes the encryption algorithm for backup phrases to AES-128-CBC with HMAC-SHA256.  The resulting ciphertext is 108 bytes when base64-encoded, making it short enough to fit into a SMS message.

The encryption scheme is specific to bip39 mnemonics.  It compresses the mnemonic to a length proportional to the number of words by translating each word into its index in the alphabetized bip39 word list.  There are 2048 words in a bip39 word list, so each word can be represented in 11 bits (3 nybbles).  This means the plaintext for any 12-word backup phrase can be safely reduced to 36 bytes without leaking information---such a phrase only has 132 bits of entropy in the first place.  The use of a HMAC is probably unnecessary since the user won't try to decrypt the ciphertext over and over, but is nevertheless included just in case this stops being true.

Nevertheless, the security of the ciphertext is derived solely from the security of the user's password---a low-entropy password will result in a ciphertext that can be brute-forced.  However, this was the case before this PR as well.  Our use of a 16-byte salt per phrase at least makes it intractable to speed up brute-force attacks with a rainbow table.

Other encryption algorithms were considered and rejected:
* AES-192-GCM.  This was rejected because to use this in Node v8, we will require an extension with a C component to add support for GCM.  Also, older versions of OpenSSL do not support GCM.  This PR does not require any new extensions.
* Vernam cipher (one-time-pad).  I went back and forth on this a few times.  On the one hand, this is potentially information-theoretically secure with a strong enough password (whereas the proposed scheme is not).  On the other hand, I don't know enough about the actual randomness of the various implementations of `crypto.randomBytes()` to know for sure that we can guarantee the OTP's randomness.  Also, ensuring that the OTP is never used anywhere else makes me nervous enough to avoid this scheme for now.
* Raw AES (no HMAC).  Typically, you want to include a HMAC with your AES ciphertext in order to prevent both ciphertext lengthening attacks and to avoid turning the user into a decryption oracle.  This is probably of lesser concern for us because the user (not a server) will be decrypting the ciphertext and we know the ciphertext length in advance. However, this may not always be true.  Also, a particularly persistent user could inadvertently help an adversary break their backup phrase by decrypting over and over.  For this reason, I included a HMAC in the ciphertext payload.
* Longer salts.  I chose a 16-byte salt to balance between having enough entropy to slow down attackers trying to brute-force the ciphertext and keeping the ciphertext short enough to send via SMS.  Making the salt longer probably does not improve security in a meaningful way, since the security of the system depends on the entropy of the user's password.
* Scrypt for key-lengthening.  I went with the standard PDKBF2 key lengthening algorithm because it's widely-deployed and "fast enough" on all target devices.  While I would have preferred Scrypt (or some other memory-hard key lengthening algorithm), I was concerned that this would hurt the user experience too much--in particular, on mobile devices.


kantai:
This looks good to me @jcnelson. For encoding/deencoding the mnemonic, though, why not just use the entropy from the mnemonic? As in:

```javascript
function normalizeMnemonic(mnemonic) {
  return bip39.mnemonicToEntropy(mnemonic)
}
```

jcnelson:
Ah, I didn't realize that method existed.  Does it do more or less the same thing?

kantai:
Not quite -- that function returns the entropy as a hex string, which the mnemonic is encoding (and, I think, according to the bip39 spec, checksums). That entropy is 16 bytes of information. I think by re-encoding the mnemonic, you end up with the checksum included (a half byte). Not sure why it ends up being 36 bytes though (is it because you're counting the bytes in the hex string?)

jcnelson:
I updated the code to use bip39 encode/decode routines.

PR #1351, opened 2018-05-11
Update Readme.md with link to installing Core

html5cat:


yknl:
Thanks for updating this, but since v0.28 a Blockstack core node is no longer required. Going to close this PR and update the readme.md file.

PR #1352, opened 2018-05-13
Fix lint in profiles/store/* and profiles/ProfilesApp

robert-bo-davis:
Fix lint profiles/store/* and profiles/ProfilesApp

contribute.blockstack.org
robert-bo-davis


yknl:
👍

PR #1436, opened 2018-05-31
Migration fixes

aulneau:
This updates the copy for the update page, along with adding a fix for if a user has an old state but no account. It will just wipe the persisted store and the next time an action fires, it will be re-saved as the latest state.

cc @yknl


aulneau:
@yknl fixed!

PR #1443, opened 2018-06-01
fixes lint for EditAccount.js

narbo:
Fixes eslint for EditAccount and removes it from eslintignore

(contribute.blockstack.org - nico)


yknl:
Thanks @narbo!

PR #1448, opened 2018-06-04
Allow 24 word keys

alvesjtiago:
- Change text when recovering account to indicate 12 and 24 word keys
- Add comparison for 24 word keys


alvesjtiago:
This PR addresses https://github.com/blockstack/blockstack-ios/issues/23

yknl:
Thanks for fixing this @alvesjtiago!

oluapocram:
Confirmed working with 0.29.2.

PR #1458, opened 2018-06-11
DRYing and simplifying

mrjmd:
This PR is some cleanup I've done while helping to familiarize myself with the codebase. It removes over 300 lines of code without changing anything of consequence.

There are a few main types of changes here across the profile components:

- Moved repetitious methods up into ``@utils`` to DRY components
- Removed long if else renders, preferring early return
- Deleted commented out code if it was 6+ months old

The only functional change should be the replacement of ``return (<div></div>)`` with ``return(<span></span>)`` for the webAccountType check in SocialAccountItem and EditSocialAccountItem. This was so that it could be combined with the this.props.listItem check.


aulneau:
Thank you for this! I will review this today and tomorrow. Appreciate the work you've put into this.

PR #1474, opened 2018-06-25
Try user's configured hub as a fail-over

kantai:
Patch to try the users gaia hub configuration when it doesn't know how to write the profile -- this will allow users who were affected by a gaia hub misconfiguration to login and use their gaia hub with applications, without requiring a zonefile update.


kantai:
Our Circle integration is currently broken, but the `npm` tests run successfully.

PR #1479, opened 2018-06-26
Use Webpack

aulneau:
This is a WIP

This PR addresses: 
- Using Webpack v4 to bundle the application
- Enabled `react-hot-loader` for faster updates (~15-30+ seconds per change to <1 second per change) in development
- Adds a `manifest.json` file for the browser itself
- A service worker is added to enable this application to be a progressive web application (add to home screen, offline, etc)
- Removes 1mb of data from initial load (from 2.5 to 1.5)
- Better image magnification and asset bundling
- Actually use `production` in production builds


hstove:
hell yeah!

aulneau:
@yknl I think this is ready for you to test out the mac build steps (should work... hopefully :))
@kantai if you have time could you also test a windows build with this branch?

Thanks!

app/public/manifest.json
```diff
@@ -0,0 +1,32 @@
...
+  "short_name": "Blockstack",
+  "start_url": "/",
+  "description": "Blockstack is a new internet for decentralized apps where users own their data.",
+  "background_color": "#ffffff",
+  "theme_color": "rgb(39, 15, 52)",
+  "display": "standalone",
+  "orientation": "portrait",
+  "icons": [
+    {
+      "src": "/images/app-blockstack.png",
```
wbobeirne: `src`'s need `/static/...`

webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+    }),
+    new workboxPlugin.GenerateSW({
+      swDest: '../sw.js',
+      clientsClaim: true,
+      skipWaiting: true,
+      modifyUrlPrefix: {
+        '/': '/static/'
+      },
+      include: [/\.html$/, /\.js$/, /\.webp/]
+    })
```
wbobeirne: I think there's a misconfig here, lacking a trailing slash on the precache manifest:
screen shot 2018-07-02 at 4 04 40 pm

webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+          name: 'vendors',
+          enforce: true,
+          test: /[\\/]node_modules[\\/]/,
+          reuseExistingChunk: true
+        }
+      }
+    }
+  },
+  plugins: [
+    new CleanWebpackPlugin(['build']),
```
wbobeirne: Should only run this if `isProd`, I might not want to clear out my build folder if I'm trying to debug something that only happens with the production bundle.

webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+      skipWaiting: true,
+      modifyUrlPrefix: {
+        '/': '/static/'
+      },
+      include: [/\.html$/, /\.js$/, /\.webp/]
+    })
+  ]
+}
+
+module.exports.serve = {
```
wbobeirne: Newer versions of webpack don't like having extra keys defined (in this case, `serve`), might as well wrap a `if (process.env...)` around this to save a future headache.

webpack.config.js
```diff
@@ -0,0 +1,196 @@
...
+  content: ['app', 'build'],
+  add: (app, middleware, options) => {
+    const historyOptions = {
+      // ... see: https://github.com/bripkens/connect-history-api-fallback#options
+    }
+    app.use(convert(history(historyOptions)))
+
+    middleware.webpack()
+    middleware.content()
+  }
```
wbobeirne: Something seems to be misconfigured here, `npm run dev` always returns a 404:

screen shot 2018-07-02 at 4 13 36 pm

package.json
```diff
@@ -136,33 +150,49 @@
     "gulp-spawn": "^0.3.0",
     "gulp-streamify": "^1.0.2",
     "gulp-util": "^3.0.6",
```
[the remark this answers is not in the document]
aulneau: Most of them can be yeah! some of the stuff we're using for testing still, haven't had the time to go in and determine which ones to keep/remove. Please feel free if you have time to start doing that, I'd love the help!

app/js/HomeScreenPage.js
```diff
@@ -20,25 +20,24 @@ function mapDispatchToProps(dispatch) {
...
-          alt={props.displayName}
-        />
+const AppIcon = ({ launchLink, iconImage, displayName }) => {
+  const webp = iconImage.replace('.png', '.webp')
+  return (
+    <div className="container-fluid app-box-wrap">
+      <a href={launchLink} target="_blank" className="app-box-container">
+        <div className="app-box">
+          <picture>
+            <source type="image/webp" srcSet={`/static/images/${webp}`} />
```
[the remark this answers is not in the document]
aulneau: Yeah I totally thought about using that too and importing the images that way, however I was weighing the cost of doing that if we plan on ultimately doing something different with the browser.

aulneau:
@wbobeirne feel free to take a look through again, I've made many of the updates you pointed out along with some others.

@yknl this is ready to be built for macos whenever you have time!

wbobeirne:
Images appear busted for me now from `npm run dev`

screen shot 2018-07-03 at 4 42 55 pm

Looking back at the code now, I see some of these image paths are dynamic. I'm not sure how well `image-webpack-loader` plays with dynamic image names, since it doesn't actually run that code. I may have steered you wrong on that recommendation, I apologize.

wbobeirne:
^ Found the fix, the font `url-loader` was interfering since they both worked on `svg` files.

aulneau:
ahh nice! thanks @wbobeirne

aulneau:
@wbobeirne since you've built the app natively, would you be able to test the mac npm tasks?

wbobeirne:
I have not actually been able to complete the build, I hit an issue I wasn't able to sort out. Sorry!

yknl:
Thanks @aulneau! I took it for a spin and found a couple of issues 

1. The nav icons are broken on the Wallet page. They are working on the Home, ID and Settings page. 
image
image

2. The production build process is crashing during `mac:copyBrowser` because the directory that it's copying to doesn't exist on a fresh build. We need to add a task to create the `./native/macos/Blockstack/Blockstack/browser` directory.

3. Error when compiling the BlockstackProxy:
`mac:compileBlockstackProxy` should be called with `npm run mac:compileBlockstackProxy`
```
✔ Entry: 'corsproxy/corsproxy.js' written to: native/macos/Blockstack/corsproxy
✔ Finished in 7.812s
sh: mac:compileBlockstackProxy: command not found
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! blockstack-browser@0.29.3 mac:release: `npm run test && npm install --no-save appdmg && npm run build && npm run mac:copyBrowser && npm run mac:compileCorsProxy && mac:compileBlockstackProxy && mac:release:build`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the blockstack-browser@0.29.3 mac:release script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

4. Should be noted that the minimum supported node version is now v8.10 due to the use of `console.debug`

These are what I found so far. There are some additional issues with the production build process that I will continue to look into this week.

wbobeirne:
Just pushed a fix for @yknl's first issue.

aulneau:
thanks @wbobeirne 

@yknl I have fixed all related issues you mentioned to the build tasks and attempted to build without a cert (or with a self-signed cert) and now I am only running into issues related to not having a correct certification to sign it with.

webpack.config.js
```diff
@@ -0,0 +1,246 @@
...
+        minifyJS: true,
+        minifyCSS: true,
+        minifyURLs: true
+      }
+    }),
+    new workboxPlugin.GenerateSW({
+      swDest: '../sw.js',
+      clientsClaim: true,
+      skipWaiting: true,
+      include: [/\.html$/, /\.js$/, /\.webp/]
```
wbobeirne: We should probably also include SVGs and fonts (woff, woff2, ttf)

aulneau: I agree, I didn't include them because we're (hopefully) removing any fonts from the browser and using system defaults

wbobeirne: OK, SVGs at least should stay since we use those as images and icons anyway. The current offline service worker driven experience has a lot of busted icons.

aulneau: sure yeah that makes sense! I also want to move away from using font-awesome and instead use [mdi-react](https://github.com/levrik/mdi-react) which I have already started doing

PR #1485, opened 2018-06-27
Updates circleCI to version 2.0

### LINKED ISSUE #1461: Migrate to CircleCI 2.0 before 2018-08-30
CircleCI 1.0 will no longer be supported after 2018-08-30

aulneau:
Circle CI will be depreciating v1 workflows in august. This PR updates the browser to use v2 of their workflow.

Resolves #1461


aulneau:
@yknl this is ready for review

yknl:
Hey @aulneau, I'm a little hesitant about switching from npm to yarn. All of our documentation currently uses npm. It's also going to use different lock file from npm which might result in a difference between local builds and circleCI builds if people are still using npm.

aulneau:
I think we should not commit either a yarn or package lock file. I've done that on other projects and it has worked out well. It allows people to use whichever they prefer. Fwiw I've been only using yarn the entire time I've worked on the browser. It's an easy switch regardless! I'm open to either, I personally think their are clear advantages of using yarn vs npm.

larrysalibra:
+1 on sticking with npm - npm is the standard package manager and it's what we're using across our project.

Agree with @aulneau that people should be able to choose their own if they want.

I'm not sure why we shouldn't commit a package lock file though. Communicating which versions of dependencies we've actually tested on seems like it's pretty important.

I actually just showed on this issue because I noticed that browser is still on circleci 1.0 and was going to open an issue if this pull request wasn't here. :-D

PR #1487, opened 2018-06-27
Default string for stringToColor

### LINKED ISSUE #1486: Profile page error due to Avatar component
The Avatar component causes the profile page to be blank because `stringToColor()` throws an exception when the `id` prop is undefined.

image

aulneau:
This gives a default value to the function in case nothing is passed in.

Should resolve #1486


yknl:
Tested and fixes #1486

Issue #1490, opened 2018-06-28
Authentication uses cached gaia config, rather than API setting


kantai:
Currently, the authentication flow tries to pass the user's custom Gaia hub URL by using a cached credential:

https://github.com/blockstack/blockstack-browser/blob/master/app/js/auth/index.js#L359

```javascript
    if (appRequestSupportsDirectHub(this.state.decodedToken.payload)) {
       hubUrl = this.props.api.gaiaHubConfig.server
    }
```

_But_, if the user changes their Gaia Hub URL setting in the Settings -> API pane, that changes:

```
api.gaiaHubURL
```

Which means that a user (even a rather determined power-user), is unable to use a custom Gaia hub. However we want to implement custom Gaia Hubs in the future (see issues #1488 #1489), this is still a bug in the behavior as I think it's supposed to work currently.

kantai:
This was merged!

Issue #1244, opened 2018-03-12
As a new user to the browser that has not added a username, I should be able to use my own gaia hub


kantai:

jcnelson:
Is this already achieved with the Gaia hub settings page?

wbobeirne:
Closing as I think Jude is right. Re-open if this is still actionable.

kantai:
There's a bug that prevents the config option from working, and there's this PR to address it: 

https://github.com/blockstack/blockstack-browser/pull/1491

PR #1491, opened 2018-06-28
Authenticate with the users configured gaia hub

kantai:
This addresses issue #1490 --

This uses the user's configured gaia hub: `settings.api.gaiaHubUrl` to set the app-specific URL and the `hubURL` field of the authentication response object.

This will allow a user who has  configured a gaia hub to use it to sign into applications (multi-player and single-player). _However_ this will not address migrating data from one hub to another -- if a user changes their gaia hub URL, they will appear to have a blank slate of data at the app-specific bucket.


aulneau:
@kantai is there a way to test that this works -- is there a different gaia hub url I can swap in without spinning up my own?

kantai:
Try this one:

`https://gaia-aaron-hub.herokuapp.com/`

jcnelson:
> This will allow a user who has configured a gaia hub to use it to sign into applications (multi-player and single-player). However this will not address migrating data from one hub to another -- if a user changes their gaia hub URL, they will appear to have a blank slate of data at the app-specific bucket.

This will also break multiplayer storage, since data written to the new Gaia hub will not be visible.  As is, this could lead to data loss and data corruption.

kantai:
> This will also break multiplayer storage, since data written to the new Gaia hub will not be visible.

No -- it will update the app key entry in the profile. Multi-player storage will continue to work.

>  As is, this could lead to data loss and data corruption.

The data is not migrated, so it will look to the user like a new login. However, it shouldn't lead to data corruption or data loss.

jcnelson:
@kantai I tested this patch.  I changed my Gaia hub API endpoint in the API settings page of the Browser, but when I signed into an app with it, the read URL was *not* changed in my profile.  I used an HTTP Gaia hub (not available via HTTPS)--maybe that interfered with updating my profile?

kantai:
@jcnelson -- I think what's going on is that the browser can't write the profile anymore once the hubUrl config changes. This patch ultimately only addresses (https://github.com/blockstack/blockstack-browser/issues/1244) -- the browser itself will have to be much more sophisticated about how hubs are configured before it will support changing the hubUrl after a name is already registered.

jcnelson:
Okay, then I think I misunderstood above.  You're saying that the Gaia hub URL in the API settings page *after I create a name* will **not** change the Gaia hub for the application I sign into (neither the read nor write paths will be affected), right?

kantai:
Changing the gaia hub url in the settings page after you've registered an ID will make the browser unable to update your profile at all.

jcnelson:
> Changing the gaia hub url in the settings page after you've registered an ID will make the browser unable to update your profile at all.

It sounds like we need a settings item for both the profile Gaia hub and an application data Gaia hub.

PR #1492, opened 2018-06-28
DPKG (`.deb) build system for Linux

kantai:
This is an oldish script I have.

We should think about shipping a deb file for the browser, rather than the docker image. This deb file basically does everything the same as the docker image does, but its 23M and doesn't require docker, so probably a better user experience overall.

Tagging @jcnelson for review, as he's the other Linux desktop user here, but this would be something that it'd be great to get community feedback on as well.

Instructions for building the deb are added to the readme.

Once it's installed, users can start a browser host process by running

`$ blockstack-browser`


dantrevino:
I used the debs primarily when they were being maintained before.  Would love to see a deb option again.

native/linux/make_deb.sh
```diff
@@ -0,0 +1,98 @@
...
+
+BUILD_DIR="$(mktemp -d)"
+
+if [ -z "$REPO_DIR" ]; then
+   REPO_DIR="."
+fi
+
+pushd $REPO_DIR
+
+VERSION=$(cat package.json | grep '"version":' | awk '{ print $2 }' | sed -e 's/"//g' | sed -e 's/,//')
```
jcnelson: Nothing wrong with this, but you might have an easier time at it if you used `jq` :)

native/linux/make_deb.sh
```diff
@@ -0,0 +1,98 @@
...
+
+echo "Copying gulped build"
+cp -r ./build $OUTPATH
+mv $OUTPATH/build $OUTPATH/browser
+
+cp native/blockstackProxy.js $OUTPATH
+
+MAKE_CORS=0
+
+if [ $MAKE_CORS == "1" ]; then
```
jcnelson: In `bash`, `=` is sufficient (`=` for equality is also POSIX-compliant).

native/linux/make_deb.sh
```diff
@@ -0,0 +1,98 @@
...
+    echo "Install the CORS Proxy"
+    mkdir -p $OUTPATH/corsproxy/node_modules
+    npm install corsproxy-https --prefix $OUTPATH/corsproxy
+    chmod -R 0755 $OUTPATH/corsproxy
+fi
+
+echo "Make run scripts"
+
+echo "#!/bin/bash" > $BINPATH/blockstack-browser
```
jcnelson: Might as well make this a here-document.

native/linux/make_deb.sh
```diff
@@ -0,0 +1,98 @@
...
+
+ARCH="all"
+NAME="blockstack-browser"
+LICENSE="MPL-2.0"
+DESCRIPTION="The Blockstack Browser"
+MAINTAINER="Aaron Blankstein (aaron@blockstack.com)"
+VENDOR="Blockstack PBC"
+URL="https://github.com/blockstack/blockstack-browser.git"
+
+fpm --force -s dir -t deb -a "$ARCH" -v "$VERSION" -n "$NAME" \
```
jcnelson: There aren't any dependencies listed here.  We should at least include `xdg-utils`, since that's what `xdg-open` lives in.  We should also include `node.js` (or whatever the package name is on Debian/Ubuntu these days).

native/linux/make_deb.sh
```diff
@@ -0,0 +1,98 @@
...
+[Desktop Entry]
+Type=Application
+Terminal=false
+Exec=blockstack-protocol-handler %u
+Name=Blockstack-Browser
+MimeType=x-scheme-handler/blockstack;
+EOF
+
+chmod 0555 $HANDLERPATH/blockstack.desktop
+chmod 0555 $BINPATH/blockstack-browser
```
jcnelson: Shouldn't `blockstack-browser` and `blockstack-protocol-handler` be executable?

native/linux/make_deb.sh
kantai: The node dependency is listed on the next line, but yeah, `xdg-utils` needs to be included as well.

native/linux/make_deb.sh
kantai: `chmod 555` indicates read + execute, right?

native/linux/make_deb.sh
kantai: Yep, will add a quote. The reason I avoided `jq` here is that this is a simple enough script so I figured it'd be better to have one less dependency in the build process.

native/linux/make_deb.sh
jcnelson: :man_facepalming: my bad

native/linux/make_deb.sh
kantai: :+1:

yknl:
@jcnelson sounds like your concerns have been addressed? Please approve the PR if so.

jcnelson:
Sorry, will do.

yknl:
Thanks!

Issue #1367, opened 2018-05-17
App private keys should use longer derivation paths


kantai:
App private keys should use a longer derivation path to take advantage of all the bits of entropy in the SHA256 hash: (see https://github.com/blockstack/blockstack-browser/blob/master/app/js/utils/account-utils.js#L55)

This PR https://github.com/blockstack/blockstack.js/pull/433 adds a function which will generate such a lengthened key, and it includes a function for generating the old derivation as well (using the `hashCode` function).

To address this in a backwards compatible fashion, we should:

1. On authentication with multi-player apps, check whether an entry _already exists_ for the app (and uses the legacy derived app key), and if so, use the legacy path, otherwise use the new path.
2. On authentication with single-player apps, check whether the app is in a whitelisted set of legacy apps (use the current set of single-player apps in the browser as the whitelist) and if so, use the legacy path, otherwise use the new path.

kantai:
This is implemented in [feature/long-app-derivation-paths](https://github.com/blockstack/blockstack-browser/tree/feature/long-app-derivation-paths) -- depends on the wallet feature branch of `blockstack.js`.

kantai:
After discussing this with @shea256, there's another approach we could take -- which is roughly:

1. Instead of using a non-secret salt, we can use a salt which is something like `ECDSA_SIGN(private_key = appsNode, message = "app-salt")`. This prevents malicious discovery of a collision, because that would require access to the salt (which they would not be able to find without the appsNode key).
2. To guard against accidental collision, we can increase the derivation path by an additional index.

The above strategy doesn't change the backwards-compatibility issues or required migration strategy, but it _would_ result in a shorter derivation path (meaning a faster key derivation).

Thoughts -- @jcnelson @larrysalibra ?

larrysalibra:
> The above strategy doesn't change the backwards-compatibility issues or required migration strategy, but it would result in a shorter derivation path (meaning a faster key derivation).

This sounds good to me.

Faster key derivation is a big plus for mobile.

jcnelson:
I'm in favor of the longer derivation path by default.  I think we have determined that derivation performance is a non-issue.

kantai:
> I'm in favor of the longer derivation path by default.

Me too -- this is from a bias towards "a new cryptographic technique is bad by default"

>  I think we have determined that derivation performance is a non-issue.

[...]

PR #1496, opened 2018-07-03
[WIP] Use long derivation paths for App Private Keys (courtesy of blockstack.js)

kantai:
This adds support for longer derivation paths for app private keys (#1367)

This maintains quasi-backwards compatibility by

1. Checking whether or not a legacy public key was used (by checking the profile's `apps` key
2. If so, using the legacy key, otherwise, uses the new derivation path.

To test this, you'll need to `npm link` in the wallet support branch on `blockstack.js` (or `develop` once it's merged).

### DIFF
--- a/app/js/auth/components/AuthModal.js
+++ b/app/js/auth/components/AuthModal.js
@@ -11,10 +11,9 @@ import {
   getAppBucketUrl, isLaterVersion
 } from 'blockstack'
 import Image from '../../components/Image'
-import { AppsNode } from '../../utils/account-utils'
+import { getCorrectAppPrivateKey } from '../../utils/account-utils'
 import { fetchProfileLocations, getDefaultProfileUrl } from '../../utils/profile-utils'
 import { getTokenFileUrlFromZoneFile } from '../../utils/zone-utils'
-import { HDNode } from 'bitcoinjs-lib'
 import { validateScopes, appRequestSupportsDirectHub } from '../utils'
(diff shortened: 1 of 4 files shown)


wbobeirne:
I've expanded this PR to use `BlockstackWallet` wherever possible.

wbobeirne:
Because `blockstack.js` has upgraded to `bitcoinjs-lib@4.x.x`, we'll need to do the same in this repo. Unfortunately that's going to be a lot of effort because all `HDNode` usage was replaced with `bip32` usage.

markmhx:
@kantai should this issue be considered resolved and closed out based on your PR? https://github.com/blockstack/blockstack-browser/pull/1620

kantai:
Nope -- as far as I understand, we still are using the short app derivation paths.

markmhx:
@kantai What's the main benefit here? Stronger security or?

kantai:
> @kantai What's the main benefit here? Stronger security or?

Yep -- this would increase the security of the app-derived keys.

markmhx:
@kantai Sounds good – it seems this PR is almost across the finish line and just needs a bit more testing. Is that the case? I'm moving this to the backlog so we can prioritize for the next sprint assuming that's the case.

hstove:
Unless I'm mistaken, this is not backwards compatible with single-player apps, because they aren't in the `apps` key. So if you login to a single-player app, your data and private key will not be the same.

PR #1504, opened 2018-07-03
Add a clear auth route for apps

### LINKED ISSUE #1502: Provide a reset state route for apps
This is paired with https://github.com/blockstack/blockstack-ios/issues/25. We'll need a new route at `/clear-auth` that presents a button that, when pressed, clears local storage. There should be a warning that makes it clear this cannot be undone. Once done, the user should be returned to the app. This route is only to be loaded from the app, so it should probably redirect to `/account/delete` if it's from a regular browsing session.

wbobeirne:
Closes #1502. Adds a simple route to `/clear-auth` that presents 2 buttons, either clear local storage or cancel. The page doesn't use the app template as the user shouldn't really do anything from this other than clear storage.

screen shot 2018-07-03 at 5 10 34 pm

### DIFF
--- a/app/js/clear-auth/index.js
+++ b/app/js/clear-auth/index.js
@@ -7,21 +7,21 @@ export default class ClearAuthPage extends PureComponent {
   }
 
   cancel() {
-    window.close()
+    window.history.back()
   }
 
   render() {
     return (
       <div className="container-fluid">
-        <h3 className="p-t-20">Clear Application Data</h3>
+        <h3 className="p-t-20">Sign Out</h3>
         <p>
           <i>
-            Are you sure you want to do this? It cannot be undone.
+            Erase your keychain and settings so you can create a new one or restore another keychain.
           </i>
         </p>
         <div className="m-t-40">
           <button className="btn btn-danger btn-block" onClick={this.clearData}>
-            Clear Data
+            Confirm
           </button>
         </div>
         <div className="m-t-10">
--- a/app/js/clear-auth/index.js
+++ b/app/js/clear-auth/index.js
@@ -1,13 +1,14 @@
 import React, { PureComponent } from 'react'
+import { withRouter } from 'react-router'
 
-export default class ClearAuthPage extends PureComponent {
-  clearData() {
+class ClearAuthPage extends PureComponent {
+  clearData = () => {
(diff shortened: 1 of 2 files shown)


hstove:
Do we need to have better logic for the redirect to `myblockstackapp://...`? Do apps use their own schemes?

wbobeirne:
@shreyasthiagaraj said the redirect should be fine, but I'll let him weigh in if there's a better method.

app/js/clear-auth/index.js
```diff
@@ -0,0 +1,35 @@
+import React, { PureComponent } from 'react'
+
+export default class ClearAuthPage extends PureComponent {
+  clearData() {
+    localStorage.clear()
+    window.location = 'myblockstackapp://?authCleared=1'
+  }
+
+  cancel() {
+    window.history.back()
```
yknl: We need to call the app's protocol handler here because going back will not close the SFAuthenticationSession.

shreyasthiagaraj: So it should be like "myblockstackapp://?authCleared=0" (where myblockstackapp is replaced with the app's provided redirect protocol)?

yknl: Yes

wbobeirne:
Thanks for the clarification, that makes a ton more sense. I've changed it to take in `protocol` query param, so you should load `/clear-auth?protocol=mycoolprotocol` to have it hit `mycoolprotocol://?authCleared={0|1}`

app/js/clear-auth/index.js
```diff
@@ -1,13 +1,14 @@
 import React, { PureComponent } from 'react'
+import { withRouter } from 'react-router'
 
-export default class ClearAuthPage extends PureComponent {
-  clearData() {
+class ClearAuthPage extends PureComponent {
+  clearData = () => {
     localStorage.clear()
-    window.location = 'myblockstackapp://?authCleared=1'
+    window.location = `${this.props.location.query.protocol}://?authCleared=1`
```
shreyasthiagaraj: Can "redirect_uri" be used instead?

wbobeirne:
@shreyasthiagaraj updated the query param

PR #1510, opened 2018-07-06
Show spinner in onboarding loader

### LINKED ISSUE #1500: Update take over loaders in onboarding/recovery steps

wbobeirne:
It looks like this was supposed to be working before, but it was a white loader on white background. Adjust the loader so it sits on top of the content instead of besides it, make it bigger, and make it blue. Closes #1500.

loader


aulneau:
Feel free to merge whenever!

PR #1511, opened 2018-07-06
Allow free subdomains outside of onboarding

### LINKED ISSUE #1499: Allow free ID registrations after sign in

wbobeirne:
Adds a check against the subdomain when registering a name outside of onboarding. Looks identical to the gif in #1204. Closes #1499.


wbobeirne:
Great feedback, just pushed a fix. You'll now only see that warning if:
1. You've attempted to register a top level name, not a subdomain
2. You have a pending top level name registration

Here's an example of it showing up after selection:
screen shot 2018-07-09 at 1 43 33 pm

yknl:
Awesome work, looks good to me now!

PR #1512, opened 2018-07-06
Fix onboarding input visual alignment

### LINKED ISSUE #1408: Visual alignment bug
Screen Shot 2018-05-25 at 10:25 PM
---
### More Info

**Source URL**: [https://www.dropbox.com/preview/Public/IMG_0120.PNG?role=personal](https://www.dropbox.com/preview/Public/IMG_0120.PNG?role=personal)

wbobeirne:
Closes #1408. This raises the text to match the labels and overlay placements (I raised the input text rather than lowered the `.id.blockstack` text because then it wouldn't have aligned with label text.) 

While I was in the neighborhood, I also converted the labels to animate on opacity and transforms, rather than using font-size, color, and top. This should perform a little better and cause less animation jank.

## Before

screen shot 2018-07-06 at 4 47 06 pm

## After

screen shot 2018-07-06 at 4 47 00 pm

### DIFF
--- a/app/js/components/ui/components/form/index.js
+++ b/app/js/components/ui/components/form/index.js
@@ -6,7 +6,7 @@ import { Form as FormikForm } from 'formik'
 const shakeAnimation = keyframes`${shake}`
 
 const Input = styled.input`
-  padding: 10px 0;
+  padding: 5px 0 15px;
   margin-top: 5px;
   border: none;
   outline: none;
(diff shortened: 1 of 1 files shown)


aulneau:
I wonder, do you think we should move down the suffix rather than add space to the bottom of the input text? Otherwise looks good to me!

wbobeirne:
That would also require moving the labels down when they're in placeholder state, since that's what the suffix was lined up with, and those would be out of alignment then. Happy to make that change if that's desired.

aulneau:
ah cool okay. I think ideally the label and the input text and the suffix would all be on the same plane, I have the preference of the original distance in the first gif. I feel like the text is just slightly too far from the line in the updated one.

wbobeirne:
Sounds good, here's the corrected spacing:

screen shot 2018-07-09 at 2 01 24 pm

aulneau:
Nice thanks!! Feel free to merge whenever!

PR #1527, opened 2018-07-17
Fix Subsequent Add ID Password Attempts

### LINKED ISSUE #1509: After password fails in create new ID modal, subsequent attempts do nothing
If you mis-type your password during the create new ID modal accessible at `/profiles/i/all`, and then try to type again, the modal simply freezes up. Attached is a gif of it happening:

create-id-freeze

wbobeirne:
Closes #1509. Refactors the the actions for creating a new ID so that they set a redux state key of `isProcessing`, rather than implicitly doing it in component state. This ensures that it's always correctly set and reusable between components, and prevents double requests at the action layer instead of the component layer.


aulneau:
Nice!!

I think when we're adding new items to state, we increment the state version to prevent bugs -> https://github.com/blockstack/blockstack-browser/blob/e822d8a161eeccab1d8baec726e69860d7d50e66/app/js/store/reducers.js#L40 but @yknl can confirm that

wbobeirne:
Good catch, the way this handled resetting isProcessing: false each time was faulty, since I don't think we quite do anything like this right now. I added a little stopgap that listens in on the `@@INIT` action instead. Hopefully cases like this can be fixed with a redux refactor, but let me know if you don't think it's satisfactory in the short term.

yknl:
Tested and looks like this one doesn't need a state version change.

PR #1528, opened 2018-07-17
Show error if manifest.json request fails

wbobeirne:
Addresses the issue in blockstack.js: https://github.com/blockstack/blockstack.js/issues/462. Right now manifest.json request failures aren't rendered at all, so a user just sees a white page. This code alone doesn't work, requires this pull request from blockstack.js to work (todo: link to pr.) Looks like this when manifest.json is rejected (Suggestions for better copy? But only devs should really see this, so it's not super important.)

screen shot 2018-07-17 at 5 39 26 pm

### DIFF
--- a/app/js/auth/index.js
+++ b/app/js/auth/index.js
@@ -48,6 +48,7 @@ import {
   selectPublicKeychain
 } from '@common/store/selectors/account'
 import { formatAppManifest } from '@common'
+import Modal from 'react-modal'
 
 const views = [Initial, LegacyGaia]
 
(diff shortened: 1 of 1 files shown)


hstove:
Can we display a different message for different errors? 404 vs 500 vs CORS error. I think it would help the developer experience a lot (particularly a CORS issue, because thats the most common)

wbobeirne:
I'm not sure how developer-y we want to get with the message since it could _potentially_ be seen by users as well. But the console is now spitting out slightly more useful information:

screen shot 2018-07-17 at 5 30 49 pm

wbobeirne:
Also blockstack.js is currently throwing a generic error, so we can't differentiate between different failures. We could expand on that, but in its current state, all errors are treated the same.

PR #1539, opened 2018-07-19
Move Mnemonic Encryption / Decryption to Web Worker

### LINKED ISSUE #1507: Move mnemonic decryption to web worker
Right now decryption blocks the main thread, causing the whole page to freeze for 2-5 seconds. Moving this over to a [web worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers) that takes in a phrase and returns a decrypted key should allow the page to stay responsive during decryption. [Support for web workers is exceptional](https://caniuse.com/#feat=webworkers), so there's little to worry about with incompatibility.

wbobeirne:
Closes #1507. Relies on #1479. Moves mnemonic encryption / decryption to a web worker so that the main thread isn't blocked while it's doing its magic. Also replaces some of the buttons that didn't have any indication of what was going on with buttons that disable and show a loader while they're working. 

It's worth noting that this PR only handles mnemonic en/decryption, and some performance issues still remain, namely with address derivation. However, I wanted to keep this initial PR pretty small to make it easier to review. But we should probably revisit the other crypto-heavy calls and workerize them.

Before:
screen shot 2018-07-19 at 4 21 34 pm

After:
screen shot 2018-07-19 at 4 21 09 pm

### DIFF
--- a/app/js/utils/encryption-utils.js
+++ b/app/js/utils/encryption-utils.js
@@ -1,117 +1,36 @@
-import bip39 from 'bip39'
-import crypto from 'crypto'
-import triplesec from 'triplesec'
+import EncryptWorker from './workers/encrypt.worker'
+import DecryptWorker from './workers/decrypt.worker'
 
-function normalizeMnemonic(mnemonic) {
-  return bip39.mnemonicToEntropy(mnemonic).toString('hex')
-} 
-
-function denormalizeMnemonic(normalizedMnemonic) {
-  return bip39.entropyToMnemonic(normalizedMnemonic)
-} 
-
-function encryptMnemonic(plaintextBuffer, password) {
-  return Promise.resolve().then(() => {
-    // must be bip39 mnemonic 
-    if (!bip39.validateMnemonic(plaintextBuffer.toString())) {
-      throw new Error('Not a valid bip39 nmemonic')
+function runWorker(WorkerClass, args) {
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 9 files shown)


aulneau:
Oooo yess look at that perf improvement! We'll have to test this in the context of native to see if it works / degrades gracefully

wbobeirne:
> We'll have to test this in the context of native to see if it works / degrades gracefully

This is only done at the browser level, not `blockstack.js`, so this wouldn't be used in node.

app/js/utils/encryption-utils.js
```diff
@@ -1,117 +1,36 @@
...
-
-function encryptMnemonic(plaintextBuffer, password) {
-  return Promise.resolve().then(() => {
-    // must be bip39 mnemonic 
-    if (!bip39.validateMnemonic(plaintextBuffer.toString())) {
-      throw new Error('Not a valid bip39 nmemonic')
+import EncryptWorker from './workers/encrypt.worker'
+import DecryptWorker from './workers/decrypt.worker'
+
+function runWorker(WorkerClass, args) {
```
wbobeirne: Having now looked at `workerize-loader` closer, it looks like it's _not_ the same as `worker-loader` and removes the need for this boilerplate. I'll give it a shot and see if the implementation is nicer.

wbobeirne:
Tests have been fixed with the help of `mock-require` to stub out a fake worker.

PR #1541, opened 2018-07-20
Add minimum length check for sponsored names 

larrysalibra:
This is the pull request corresponding the length check in the subdomain registrar: https://github.com/blockstack/subdomain-registrar/pull/16

It checks to make sure usernames are at least 8 characters.


jeffdomke:
Looked at this quick, seems like we need to: 
– Tell user the min length upfront. 
– Have the error in place (the space or those errors is a teeny, tiny area for very short messages. 
– Show not available (not sure if that is already covered)

PR #1554, opened 2018-07-25
Onboarding Stats

### LINKED ISSUE #1546: Investigate metrics solution: https://matomo.org/ or extending our mixpanel approach for the browser.

wbobeirne:
### Description

Closes #1546. Relies on https://github.com/blockstackpbc/blockstack-browser-server/pull/14. Adds analytics to onboarding by tracking each step for sign up / sign in

### Changes

* Adds a server API class used to do all calls to `browser-api.blockstack.org` instead of one-off fetch calls.
* Adds two functions, `trackEvent` and `trackEventOnce` for submitting event tracking requests.
* Adds two keys to state.settings.api, `distinctEventId` and `hasDisabledEventTracking`.
    * `distinctEventId` is used to disambiguate events and is persisted across all sessions
    * `hasDisabledEventTracking` is a boolean to disable tracking. The option to do so has been added to settings.
* Refactors the store to be importable outside of the Redux provider for libraries to use.
* Refactors local storage usage to be synchronous

### Screenshots

screen shot 2018-07-25 at 2 26 48 pm


app/js/sign-in/index.js
```diff
@@ -112,12 +119,14 @@ class SignIn extends React.Component {
     }
   }
 
+  componentDidMount() {
+    trackEventOnce(VIEW_EVENTS[this.state.view])
```
wbobeirne: Add referring app as a property.

app/js/utils/server-utils.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+export function trackEvent(event: string, properties: Object) {
+  const state = store.getState()
+
+  if (state.settings.api.hasDisabledEventTracking) {
+    return
+  }
+
+  // Prevent noisy stats from development testing behavior
+  if (process.env.NODE_ENV !== 'production') {
```
yknl: Looks like `process.env.NODE_ENV` is `undefined` here in both dev and prod.

wbobeirne: A quick search shows we use it all over the codebase, and the gulp tasks should be setting it: https://github.com/blockstack/blockstack-browser/search?q=process.env.NODE_ENV&unscoped_q=process.env.NODE_ENV

Any idea why that might be?

yknl: Seems to be an existing bug that NODE_ENV is undefined. @wbobeirne Are you getting this as well?

PR #1555, opened 2018-07-25
Move Email to End of Sign-in

### LINKED ISSUE #1551: On restore move email entry step to after password entry. When password fails and user restarts, this ads one more step they must repeat.

wbobeirne:
### Description

Closes #1551. Moves the email step to after password. This required a little reworking of the logic, and as a result, I attempted to solve a lot of the race conditions that were using `setTimeout` to get around.

### Steps to Test

1. Start a fresh instance of the browser
2. Recover using a magic recovery code
3. Confirm the order is correct, and recovery was succesful
4. Wipe browser again and repeat with mnemonic phrase

### DIFF
--- a/app/js/sign-in/index.js
+++ b/app/js/sign-in/index.js
@@ -44,12 +44,12 @@ const logger = log4js.getLogger('sign-in/index.js')
 
 const VIEWS = {
   INITIAL: 0,
-  EMAIL: 1,
-  PASSWORD: 2,
+  PASSWORD: 1,
+  EMAIL: 2,
   SUCCESS: 3
 }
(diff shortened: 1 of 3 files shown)


yknl:
@wbobeirne we have a regression in this PR. After restoring a confirmed Blockstack ID, the identity refresh is not being fired so it displays Nameless User. 

image

wbobeirne:
Got it, I'll see if I can restore that behavior before release.

wbobeirne:
Great catch, fixed in https://github.com/blockstack/blockstack-browser/pull/1569

Issue #368, opened 2017-04-03
Expand browser test coverage


larrysalibra:
Now that we have a working test system & CI, we should expand our test coverage for existing code.

wbobeirne:
Closing this in favor of individual issues around testing that have clearer closing criteria.

PR #1561, opened 2018-07-29
Add more component tests

dashaxiong:
This PR includes more unit tests for app/js/components. It helps with issue #368. 

The tests are based on existing code, testing some of the ways that components are rendered when supplied with different props. They typically check if the render function produces particular elements with specific attributes and text nodes. Very happy to discuss ways in which the tests may be improved :hammer: :grin:  

New test suites are added for components that were previously untested.

### DIFF
--- a/test/components/Navbar.test.js
+++ b/test/components/Navbar.test.js
@@ -12,99 +12,48 @@ describe('<Navbar />', () => {
     })
   })
 
-  describe('renders the component with active home tab', () => {
-    const props = {
-      activeTab: 'home'
-    } 
+  const tabDetails = [
+    {
+      name: 'home',
+      activeIcon: '/images/icon-nav-home-hover.svg',
+      regularIcon: '/images/icon-nav-home.svg'}, 
+    {
+      name:'settings',
+      activeIcon: '/images/icon-nav-settings-hover.svg',
+      regularIcon: '/images/icon-nav-settings.svg'}, 
+    {
+      name:'wallet',
+      activeIcon: '/images/icon-nav-wallet-hover.svg',
+      regularIcon: '/images/icon-nav-wallet.svg' }, 
+    {
+      name: 'profile', 
+      prop: 'avatar',
+      activeIcon: '/images/icon-nav-profile-hover.svg',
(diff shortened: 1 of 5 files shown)


test/components/Navigation.test.js
```diff
@@ -0,0 +1,35 @@
...
+    }  
+  
+    const wrapper = shallow(<Navigation {...props} />)
+
+    it('uses a ChevronLeftIcon', () => {
+      expect(wrapper.find('ChevronLeftIcon').exists()).to.equal(true)
+    })
+
+    it('includes the previous label', () => {
+      expect(wrapper.childAt(0).childAt(1).text()).to.contain('Back')
```
wbobeirne: This is really brittle, selecting specific child indices. It should probably search for an element by tag, or simply check that the text is contained.

test/components/Navigation.test.js
```diff
@@ -0,0 +1,35 @@
...
+  describe('renders the component with a label', () => {
+    const props = {
+      previous() {},
+      previousLabel: 'Previous Page'
+    }
+
+    const wrapper = shallow(<Navigation {...props} />)
+
+    it('includes the supplied label', () => {
+      expect(wrapper.childAt(0).childAt(1).text()).to.contain('Previous Page')
```
wbobeirne: Same as above.

test/components/Navbar.test.js
```diff
@@ -0,0 +1,110 @@
...
+    })
+
+    it('uses the regular wallet tab icon', () => {
+      expect(wrapper.find('img[src="/images/icon-nav-wallet.svg"]').exists()).to.equal(true)
+    })
+
+    it('uses the active profile tab icon', () => {
+      expect(wrapper.find('img[src="/images/icon-nav-profile-hover.svg"]').exists()).to.equal(true)
+    })
+  })
```
wbobeirne: I'd like to see less copy-paste of these tests. Looping over an array of possibilities would make for a more easily maintained test if something about Navbar changes.

dashaxiong: Thanks for your feedback, I agree. I will attempt those changes today.

PR #1565, opened 2018-08-03
Update non-use css class in NotFoundPage.js

kaxing:
Notice that 404, Page Not Found warning is mostly covered under "Navbar"*, did quick check and found out "app-wrap-profiles" is not being defined. If I understand correctly, using "body-main" would be proper in this one.

* as seen in Safari image


wbobeirne:
Hey @kaxing, thanks for the contribution! I actually _just_ fixed this one with #1559 about the same time you submitted, so this has been resolved. Thanks again though for pitching in.

Issue #1547, opened 2018-07-23
Testing on nameless users


jeffdomke:

markmhx:
@jeffdomke I'm not really sure how to interpret this issue. Want to revise it a bit for clarity, or best to close out given its age?

moxiegirl:
@markmhx I think Pete one of evangelists is running into this issue --- hard to tell from the description what prompted it.  He posted in channel Wednesday.

markmhx:
@moxiegirl Got a link handy to his post? I'll take a look. We may want to create a new issue that maps more directly and clearly to his concerns since this one is pretty old and sparse.

moxiegirl:
@markmhx https://blockstack.slack.com/archives/GADLAAFGB/p1552057323032700

markmhx:
@moxiegirl Hmm I'm unable to view that message for some reason, mind checking the link?

<img width="647" alt="Screen Shot 2019-03-12 at 10 54 59" src="https://user-images.githubusercontent.com/28991/54191026-5fe31e00-44b5-11e9-8b26-654fe5629742.png">

moxiegirl:
@markmhx worked for me. the link is to Evangelist official in the open Slack. Looks like you were not in that channel. Invite sent

PR #1583, opened 2018-08-07
Make Onboarding More Failure-Resistant

### LINKED ISSUE #1506: Onboarding doesn't handle emails that fail to send
When signing up for a new account, requests are sent to browser-api.blockstack.org. These requests could fail for various reasons (User is offline, api rejects the request, one of the down-stream apis fails.) However, even when this happens, onboarding just continues with the screen letting them know they've received an email, even if they never will.

The user should probably instead either see (a) some kind of in-browser version of the backup information, maybe something they can write down or print, and / or (b) a screen to let them know the request errored out, and to try again later.

Here's the relevant requests in code, they should be behaving differently in the `catch` clause, and should be handling if the response is `success: false`: https://github.com/blockstack/blockstack-browser/blob/master/app/js/sign-up/index.js#L377 & https://github.com/blockstack/blockstack-browser/blob/master/app/js/sign-up/index.js#L408

screen shot 2018-07-05 at 11 11 48 am

### LINKED ISSUE #1524: Onboarding doesn't handle subdomain registration errors
Errors can be returned by the subdomain registrar. Currently the onboarding flow simply ignores those errors and continues onboarding the user. For example:

* A user may have signed up for too many names already with a given IP.
* The submitted username may have invalid characters -- it seems this happens on mobile, because the mobile keyboard autocorrects the name to capitalized
* The server may be broken

Those errors should be manifested to the end user.

### LINKED ISSUE #1548: Lower timeout for network requests and add useful error messages

wbobeirne:
Closes #1506. Closes #1524. Closes #1548.

### What This Does

This adds failure handling to most aspects of onboarding. You'll now both be informed of, and will be able to continue through onboarding bug-free if the following happen:

* You're unable to check the availability of your username
* Your username registration fails
* Your recovery email fails to send

Email requests are now done in parallel, and cap out at 10 seconds on slow connections.

An implication of this change is that users can now end up in the browser after onboarding without a username. While this was already a use case we supported, it'll be more important now that we behave without issue in that case (Hopefully we suss out all issues after #1547 is complete.)

In addition to the above features and changes, this PR adds the [`reapop` library](https://github.com/LouisBarranqueiro/reapop) as per @aulneau's suggestion for showing error messages. This should be useful in other parts of the browser. Here's an example of how to use:

```.jsx
import { notify } from 'reapop'

class MyComponent extends React.Component {
  componentDidMount() {
    this.props.notify({
      title: 'Hello!',
      // ...
    })
  }
}

connect(mapStateToProps, () => {
  notify
})(MyComponent)
```

### Steps to Test

The easiest way to test all of these cases is to use [Chrome's "Request Blocking" feature](https://developers.google.com/web/updates/2017/04/devtools-release-notes#block-requests).

* **Failed Username Check**
  1. Go through onboarding until the username step
  2. Block requests to `registrar.blockstack.org`
  3. Attempt to check on a username
  4. Confirm you get an error and the skip button appears
  5. Skip the step, confirm onboarding continues correctly and shows no username at the end
* **Failed Username Registration**
  1. Go through onboarding until the username step
  2. Attempt to check on a username succesfully
  3. Block requests to `registrar.blockstack.org`
  4. Attempt to register the username
  5. Confirm that after it finishes, you see a popup informing you it failed
  6. Confirm onboarding continues correctly and shows no username at the end
* **Emails failed to send**
  1. Go through onboarding until the username step
  2. Block requests to `browser-api.blockstack.org`
  3. Complete the step, confirm that the next step you see informs you of the failure to send the email
  4. Unblock requests and click "resend"
  5. Confirm that the email successfully resends
  6. Confirm onboarding continues correctly

### Failed to check username availability (w/ skip button)

screen shot 2018-08-07 at 3 01 15 pm

### Failed to register username

screen shot 2018-08-07 at 3 02 23 pm

### Email failed to send

screen shot 2018-08-07 at 3 04 07 pm

### End screen after failed / skipped username registration

screen shot 2018-08-07 at 2 49 15 pm


app/js/components/ui/containers/user.js
```diff
@@ -33,7 +33,7 @@ const UserAvatar = ({
         }}
       />
     ) : (
-      <span>{firstLetter(username)}</span>
+      <span>{firstLetter(username || '?')}</span>
```
wbobeirne: To clarify this change, this makes sure this works even if you pass `null` or `''`.

app/js/sign-up/index.js
```diff
@@ -312,25 +332,35 @@ class Onboarding extends React.Component {
...
+    let recoveryPromise = Promise.resolve()
+    let restorePromise = Promise.resolve()
+    if (type === 'recovery' || type === 'both') {
+      recoveryPromise = sendRecoveryEmail(email, id, encodedPhrase)
+        .then(() => this.setState({ recoveryEmailError: null }))
+        .catch((err) => this.setState({ recoveryEmailError: err }))
+    } if (type === 'restore' || type === 'both') {
+      restorePromise = sendRestoreEmail(email, id, encodedPhrase)
+        .then(() => this.setState({ restoreEmailError: null }))
+        .catch((err) => this.setState({ restoreEmailError: err }))
```
wbobeirne: Side note, can we just combine these two emails into one? Getting two emails immediately is kinda weird.

app/js/sign-up/index.js
```diff
@@ -250,16 +251,34 @@ class Onboarding extends React.Component {
...
         address,
         keypair
       )
-      this.setState({
-        creatingAccountStatus: CREATE_ACCOUNT_SUCCESS
+      .catch((err) => {
+        logger.error(`username registration error: ${err}`)
+        this.props.notify({
+          title: `Failed to register ${username}`,
+          message: 'Something went wrong while registering that name. You ' +
```
moxiegirl: > Something went wrong while registering that name.

"that name" is an odd construction.

title: `Name registration failed`,
message: 'Something went wrong while registering ${username}. You ' +

jeffdomke:
This is super cool. My main questions are: 
- When we skip the name, what failures get introduced down the line? Stealthy won't work for example. 
- I'm assuming hitting continue after the email failure step forces the user to record the seed (they only have one option opposed to two)?

wbobeirne:
> When we skip the name, what failures get introduced down the line? Stealthy won't work for example.

Legacy users today already don't have names, so users who fail here would get the same experience as those users. If we know certain apps won't work without usernames (Is it a particular permission like publish data that doesn't work?) we should probably stop the user from authing the app until the user's added a name for themselves.

> I'm assuming hitting continue after the email failure step forces the user to record the seed (they only have one option opposed to two)?

No, but that seems like a good idea. I can change it to be like that if you think that's the way to go.

yknl:
> When we skip the name, what failures get introduced down the line? Stealthy won't work for example.

I think the app should let the user know if they need a username. Stealthy currently does this.

jeffdomke:
Re Ken's comment. What other options to do we have? 
- Force them to retry. 
- Let them skip and include a warning that says "You can skip picking a username but some apps won't behave properly." 

Anything else?

wbobeirne:
> Force them to retry.

I think this isn't an option. If the onboarding _requires_ these outside network requests, then it's no longer failure-proof and hitched to DNS / our sponsored domain registrar / being online. The point of these changes is to allow a user to complete onboarding if all else fails.

> Let them skip and include a warning that says "You can skip picking a username but some apps won't behave properly."

We could definitely spice up the language around skipping and failures to make this clear. However, it would be great to understand _why_ some apps require names, and make them be more explicit about it during auth or figure out how we can detect it. That way we could prevent a user from every getting a broken experience by stopping them at the auth step and telling them they need a name to continue.

yknl:
> We could definitely spice up the language around skipping and failures to make this clear. However, it would be great to understand why some apps require names, and make them be more explicit about it during auth or figure out how we can detect it. That way we could prevent a user from every getting a broken experience by stopping them at the auth step and telling them they need a name to continue.

Multiplayer apps like Stealthy need usernames so user's can find each other.

jeffdomke:
> We could definitely spice up the language around skipping and failures to make this clear. However, it would be great to understand why some apps require names, and make them be more explicit about it during auth or figure out how we can detect it. That way we could prevent a user from every getting a broken experience by stopping them at the auth step and telling them they need a name to continue.

Agree, I'm assuming detection is tricky and would be a later enhancement? So for this immediate release we just mention the _why_ and basically apologize?

wbobeirne:
> Multiplayer apps like Stealthy need usernames so user's can find each other.

If that's the case, let's just block nameless users from auth'ing apps with `publish_data` as a permission. The auth screen will tell them to register a name instead.

> So for this immediate release we just mention the why and basically apologize?

Sure, so the changes I see for that are:
* If checking name availability fails, add some additional copy below the name input that says something like "Sorry, we're having an issue registering a name. You can try again or skip this and do it later. This may make some apps unusable until you do."
* If name registration fails when you try to register, show the error popup with something like "Sorry, something went wrong with registering [yourname].id.blockstack. You can try to register again later from your profile. Some apps may be unusable until you do."

jeffdomke:
Sounds right to me.

wbobeirne:
Here's what happens now if check username fails:

screen shot 2018-08-14 at 3 08 42 pm

^ The second line of copy only pops up after failure, otherwise it looks identical.

And here's what the registration failure message says now:

screen shot 2018-08-14 at 3 11 36 pm

yknl:
It was able to catch the 409 error, but the recovery emails are broken due to the lack of a username.
Looks like we need to modify the emails so that they can handle the case of no username.

image

wbobeirne:
Sorry for leaving this one hanging for a bit, totally forgot about the emails. Is there a good way to address users who don't have a blockstack name? We use "Nameless user" in the onboarding UI, but that seems very unfriendly for an email.

yknl:
I think we should just avoid addressing them in the emails if there's no username.

wbobeirne:
To make this easier to reason about, here's all of the spots we use blockstack id in the copy for emails:

Seed email:

> ${blockstackId} - Action required: Record your Secret Recovery Key
> ---
> #{blockstackId} - Record your Secret Recovery Key
> ⚠️ Action required: Record your Secret Recovery Key for #{blockstackId}

Recovery email:

> ${blockstackId} - Save this email: Magic Recovery Code
> ---
> #{blockstackId} - Save this email: Magic Recovery Code
> Save this email: Magic Recovery Code for #{blockstackId}
> Magic Recovery Code for #{blockstackId}: #{encryptedSeed}

yknl:
Proposal:

> ## Action required: Record your Secret Recovery Key
> Record your Secret Recovery Key
> ⚠️ Action required: Record your Secret Recovery Key 

> ## Save this email: Magic Recovery Code
> Save this email: Magic Recovery Code
> Your Magic Recovery Code: #{encryptedSeed}

wbobeirne:
^ Donezo, see it at https://github.com/blockstackpbc/blockstack-browser-server/pull/15

PR #1606, opened 2018-08-13
Unify Store Config

### LINKED ISSUE #1603: Confirm redux stores work the same in dev + prod environments
With the merge of the webpack PR we will now be releasing production code whereas before we were only shipping development code. There are some differences for the redux store between dev and prod, we should confirm there are no regressions from shipping prod code.

wbobeirne:
Closes #1603. Moves redux store configuration into a single file so that redux behaves the same in dev as in prod. Previously it wasn't clearing the auth reducer state on every session.


yknl:
merging this in as a hotfix

PR #1611, opened 2018-08-14
Add better loader before js is parsed and executed

aulneau:
This PR adds a simpler loader for when the browser is parsing the js. As soon as the client loads up it will add a `hidden` class to the loader and animate away. This helps improve our lighthouse score.

You can see the loader here -> https://file-ephxprodgz.now.sh/

~~This PR also adds in preloading of our bundles after the performance update was merged. With that PR the routes were all split into their own bundles, this will prefetch them after the client has loaded.~~

Our service worker will fetch all of our assets after initial load :)

Lighthouse score with this PR:
screen shot 2018-08-14 at 3 01 43 pm


app/public/index.html
```diff
@@ -32,33 +32,47 @@
...
         margin: 0;
         padding: 0;
-
+        top: 0;
+        left: 0;
+        position: absolute;
+        pointer-events: none;
+        transition: 0.25s all ease-in-out;
+        z-index: 99999999999999;
+        transform: translate3d(0, 0, 0);
```
wbobeirne: Opinion: I think it would be nice if the SVG animated fading in and out a bit, just so the user understands that the page is working and hasn't crashed or anything like that.

aulneau:
@wbobeirne my understanding is that because they are links and because of `rel="preload"` they are non-blocking. -> https://www.smashingmagazine.com/2016/02/preload-what-is-it-good-for/

> On top of that, preload does not block the window’s onload event, unless the resource is also requested by a resource that blocks that event.

aulneau:
Okay I've added an animation and adjusted it from being 'preload' to 'prefetch' which should give us the initial load perf we hope for while also prefetching all the other assets after initial load. cc @wbobeirne

yknl:
Nice, really love the animating Blockstack logo.

PR #1614, opened 2018-08-15
Improve logging accuracy & display

wbobeirne:
Slight improvements to our logging:

* Uses styling to de-emphasize the path of logging
* Uses `__filename` instead of hard coded string to get file context. Should fix a bunch of incorrect paths.
* Replaces all `logger.trace` calls with `logger.info`. You can still use `logger.trace` to print stack traces, but most uses before were not written with that intent and it spammed the console pretty bad.

Here's what the new styles look like:

screen shot 2018-08-15 at 3 51 17 pm

### DIFF
--- a/app/js/App.js
+++ b/app/js/App.js
@@ -22,7 +22,7 @@ import { hot } from 'react-hot-loader'
 
 import log4js from 'log4js'
 
-const logger = log4js.getLogger('App.js')
+const logger = log4js.getLogger(__filename)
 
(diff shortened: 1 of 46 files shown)


wbobeirne:
Looks like a bunch of our tests actually assert logging messages. Is this something we want to keep in? Honestly that kind of seems a little overkill to me.

yknl:
I'm down with removing them.

Issue #868, opened 2017-10-08
Improve wallet input validation and error handling


larrysalibra:
Right now we just return error messages from core.

PR #1619, opened 2018-08-17
Transaction Confirmation Modal

wbobeirne:
This fixes a bunch of UX issues, and potentially some bugs with the wallet. Not tracked by an issue.

### What This Does

This PR splits transactions into two actions, signing and broadcasting. Signing decrypts the user's private key, passes off the details to `blockstack.transactions.makeBitcoinSpend`, and sets the raw transaction hex to the store. The user is then prompted with a modal to confirm the transaction as decoded from the raw hex. This gives them the opportunity to see the amount sent with the fee subtracted.

To work within the limits of blockstack.js, this still retains the behavior that the amount you type in is not the amount the desired address will receive. We may want to add some language around this, either in the form or in the confirmation modal.

This also adds better loading behavior, and better error handling for amounts that are too large or too small.

### Steps to Test

* Get on an account with bitcoin in the wallet
* Attempt to send more than the balance, confirm you get an error
* Try to send an amount less than `1e-8`, e.g. `0.000000000001`, confirm you get an error
* Click the "send all" button, confirm it puts in the whole balance
* Submit the form, confirm you get a confirmation modal
* Send the transaction, confirm you get a success message
* Confirm the transaction was actually broadcast by checking your address on an explorer

### Demo

wallet-demo


yknl:
Works great! Thanks @wbobeirne.

Can we make the error message below more clear? It sounds like I don't have enough in my balance to cover the transaction fee when I do. Maybe also suggest that they enter a higher spend amount to cover the transaction fees. 

image

wbobeirne:
Looks like those messages come from Core (Issue tracked in #868.) I don't know that we have a way to disambiguate them. What would you suggest we do?

PR #1631, opened 2018-08-22
Change Travelstack information

alvesjtiago:
This commit changes 3 properties of Travelstack reference:
- launch link to app url;
- storage required set to true;
- status as user ready.


wbobeirne:
Thanks for cleaning this up!

PR #1633, opened 2018-08-23
Fix loader on clear-auth

### LINKED ISSUE #1632: Clear auth page loads indefinitely
@petec029 reported this: https://github.com/blockstack/blockstack-ios/issues/50#issuecomment-415018182

I have the same issue when I navigate to https://browser.blockstack.org/clear-auth on desktop Chrome or iphone.

wbobeirne:
Closes #1632. Wraps pages that weren't wrapped in `<App />` so that they get the same behavior as other pages, which includes dismissing the loader.

### Steps to Test

* Load `/clear-auth`, confirm it shows.

### DIFF
--- a/app/js/App.js
+++ b/app/js/App.js
@@ -55,9 +55,12 @@ function mapDispatchToProps(dispatch) {
 
 class AppContainer extends Component {
   static propTypes = {
+    // Own props
+    children: PropTypes.element.isRequired,
+    noHeader: PropTypes.bool,
+    // State props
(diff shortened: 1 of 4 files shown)


app/js/App.js
```diff
@@ -183,6 +186,7 @@ class AppContainer extends Component {
 
   componentDidMount() {
     const loader = document.getElementById('loader')
+    console.log(loader)
```
wbobeirne: Whoops, removing

PR #1634, opened 2018-08-27
Feature/consistently order social proofs revised

craigiansmith:
This is a revised branch which sorts the social media account proofs in the Default Profile Page in consistent alphabetical order. It includes tests for the Default Profile Page component to make sure that the accounts are consistently ordered when they are unfilled, and when some accounts are filled. It ensures that filled accounts precede the unfilled accounts.

### DIFF
--- a/app/js/profiles/DefaultProfilePage.js
+++ b/app/js/profiles/DefaultProfilePage.js
@@ -70,7 +70,8 @@ function mapDispatchToProps(dispatch) {
   )
 }
 
-class DefaultProfilePage extends Component {
+// Named export to test the Component in isolation
+export class DefaultProfilePage extends Component {
   static propTypes = {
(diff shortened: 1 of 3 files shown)


app/js/profiles/DefaultProfilePage.js
```diff
@@ -70,7 +70,8 @@ function mapDispatchToProps(dispatch) {
   )
 }
 
-class DefaultProfilePage extends Component {
+// Named export to test the Component in isolation
+export class DefaultProfilePage extends Component {
```
craigiansmith: I have added a named export to the DefaultProfilePage component to allow the unconnected component to be unit tested as per the [ReduxJS Writing Tests guide](https://github.com/reduxjs/redux/blob/master/docs/recipes/WritingTests.md#connected-components)

Issue #1575, opened 2018-08-06
Research testing processes and come up with a testing script.


yknl:

wbobeirne:
Just dumping my personal experiences here

### Automated Cross Browser Testing

In general, this stuff is extremely brittle and requires a ton of effort from engineers to maintain its functionality. However, it can catch a bunch of useful stuff once well tuned. Some options I've had limited experience with:

* https://www.seleniumhq.org/
* https://www.cypress.io/ 

### Headless Browser Testing

Much easier to develop, keep up to date, and run are headless browser tests. However, typically only works for Chromium or Firefox since it's something the browser has to build support for.

* Traditionally I'd used PhantomJS or SpookyJS, but both projects have since dropped maintenance. I don't know what the hot newness is here.

### Cross Browser Smoke Testing

Often a good starting point, simply load a few things in every browser and make sure it doesn't explode. It doesn't capture most (or really any) of our complicated workflows, such as authentication, but it's better than nothing.

We can also get our mocha tests to run in each browser, which ensures that all browsers support the APIs we leverage, and the babel-transpiled code. And you basically get it for free, since we already have a bunch of tests.

* https://mochajs.org/#running-mocha-in-the-browser run in BrowserStack or SauceLabs

### Testing Scripts

Regardless of whether it's manual or automated, we should have human-language scripts to be either implementation specs, or guides for people to read on how to test features. Many projects have these, but I personally helped work on [MyCrypto's Workflow Checklists](https://github.com/MyCryptoHQ/MyCrypto/wiki/Workflow-Checklists) which are pretty comprehensive, especially for some of the really niche smart contract features.

aulneau:
I haven't used it personally, but have heard very positive things about cypress.io

markmhx:
@zone117x is this issue still relevant given your recent work on testing?

zone117x:
I'd consider this closed

Issue #1481, opened 2018-06-26
Explore other testing frameworks


aulneau:
I've heard very good things about https://www.cypress.io/, I'd be curious to explore it as an option for End to End testing of the browser.

craigiansmith:
@alneau I tried out cypress and ran into this [issue](https://github.com/cypress-io/cypress/issues/1526) as it uses an older library on Ubuntu which causes it to throw Critical warnings. Could still be useful though. 
Perhaps [SuperAgent](https://visionmedia.github.io/superagent/) is worth looking into also.

markmhx:
@aulneau should we close this out since it's more of a research task than anything?

cc @zone117x since he's been working on testing since this issue was created, in case it interests him

zone117x:
So far, we have implemented integration tests that have the requirement of testing against a matrix of different web browsers, operating systems, and devices. Given this requirement, the only testing frameworks capable are ones that use Selenium/WebDriver. BrowserStack [lists about all of them here](https://www.browserstack.com/automate/node). Our current e2e tests in this repo use the [lib](https://www.npmjs.com/package/selenium-webdriver) maintained by the Selenium people. 

The downside is that the Selenium stack is horribly slow and buggy (regardless of which wrapper lib used). It's difficult and time consuming to write and update test suites. 

Additionally, there are certain types of integration tests we need in both blockstack.js and this repo which have contradicting requirements, and are infeasible to write using the Selenium stack. 

For example:
* @hstove has brought up testing our React code with [Jest](https://jestjs.io/) which doesn't need to be ran in a browser environment. 
* blockstack.js needs integration tests implement with the following requirements:
  * A large quantity of small tests. 
  * Ran in a browser environment, but does not need cross-platform, so fast headless Chromium-based frameworks like cypress.io and https://github.com/GoogleChrome/puppeteer would work. 

I added the `[QA]` prefix in the issue title as this is something our new QA engineer should probably take on. @markmhx have we decided how these QA issues should be labeled or categorized?

markmhx:
Sounds good re: expecting a new QA engineer to take on these considerations.

We haven't decided anything new about labels per se, but meanwhile, I'd encourage us to add all QA-related ones to this epic: https://github.com/blockstack/blockstack-browser/issues/1862

PR #1642, opened 2018-09-03
Add Cypress.io for running end-to-end integration tests

craigiansmith:
Addresses #1575, and #1481

**This PR adds Cypress as a dependency and creates an initial test suite for the sign up workflow.**

To install it:  `npm install`

Then on Ubuntu, you will also need to install libgconf-2.4, which is an outdated dependency which throws Glib-GIO-CRITICAL warnings as mentioned in this [issue](https://github.com/cypress-io/cypress/issues/1526). Install it with: `sudo apt install libgconf-2-4`.

The integration test is located at _test/cypress/integration_, though this can be configured differently. Out-of-the-box it uses a _cypress_ directory in the root of the project. Occassionally, cypress creates that folder, even though new locations have been specified in _cypress.json_.

To prepare the test environment:
`npm run dev`

Then when the server is built:
`npx cypress open`

This will launch a new window, click: `run all specs` and it will start running through the tests.

To run the tests headless, use `npx cypress run`.

What do you think :question: 
     
@aulneau @yknl @wbobeirne


craigiansmith:
The extra cypress folder is no longer created. There was a typo in the config file, which is now fixed.

yknl:
Hey @craigiansmith, we've start implementing end-to-end tests using selenium-webdriver on BrowserStack. We evaluated cypress.io but found that it doesn't suit our needs. Our goal with automated end-to-end tests was to test basic functionality across multiple browsers/platforms. Cypress.io doesn't have support for anything other than chrome-like browsers.

aulneau:
Thanks for your work on this @craigiansmith 

I would perhaps reconsider using this @yknl, related -> https://www.cypress.io/blog/2018/09/20/cypress-is-officially-out-of-beta/

Coming soon:
> Cross browser support with native events, Sauce Labs, and even IE11.

I've heard a lot of the internet really praise cypress (I don't have any personal experience with it).

yknl:
I think we can reconsider cypress once their cross browser support is released. And if mobile browsers are also supported.

yknl:
Closing this for now as there's no cross browser support in cypress right now.

PR #1668, opened 2018-09-17
Some minor webpack fixes

aulneau:
This PR addresses the following:

- Added [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin) in place of uglify (is faster, performs the same if not better and is actively maintained)
- Added [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) so we can easily monitor our bundle sizes and reduce bloat where we cant.
- Made some updates to our chunking process for better longterm caching (uses contenthash now, and each module gets a name, generally) [see these lines](https://github.com/blockstack/blockstack-browser/pull/1668/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR196)
- Added webpack-stylish for cleaner webpack output
- Updated some deps
- Our service worker was not saving to the correct path, now it is and now we'll be using our service worker again :)
- ~~we're using `webpack-serve` now instead of the no longer maintained `webpack-dev-server`~~ EDIT: ok so this was depreciated yesterday... moving back to webpack-dev-server :(- removed the import of the entire `mdi-react` lib into our vendors file, now we're importing just what we are using (cut our vendors.bundle in half)
- Cleaned up the production output significantly, everything is now outputted to the correct places (eg `static/js/*`)


aulneau:
@yknl this should be good for you to review whenever you have time!

yknl:
I can't seem to be able to run in development with these changes. Starting the browser in an incognito window results in the loading graphic showing indefinitely without errors logged to console. I used node v8.11.2.

aulneau:
Okay that was strange I think it broke because I bumped the 'file-loader' dep up. Should be good now, tested and working. I just did a couple more quick fixes:

- we're using `webpack-serve` now instead of the no longer maintained `webpack-dev-server`
- removed the import of the entire `mdi-react` lib into our vendors file, now we're importing just what we are using (cut our vendors.bundle in half)

aulneau:
@yknl ok merged from develop and fixed conflicts, should be good to go again :)

aulneau:
Nice thanks! merging now :)

PR #1674, opened 2018-09-19
[WIP] Onboarding ui and copy updates

aulneau:
This [WIP] PR addresses the following:

- Users will no longer lose the value of recover code input if they go back (eg Recover Code -> Password -> Back to recovery code)
- Updated to styled-components v4
- Updated some of the async functions for restoring an account
- Cleaned up the design of the profile view when restoring account
- Reduced spacing of inputs
- Hides labels on focus
- Refined when back buttons show up
- Most onboarding screens will/should have only one cta (other than back)
- ...
- more soon!


yknl:
Getting a blank page and an error clicking into sign in or restore on the latest commit

image

aulneau:
@yknl looking into it...

yknl:
Looks like there's an error with the final profile screen if you go through the restore flow, which results in a blank screen. 
image

aulneau:
@yknl whoops thank you, fixed with https://github.com/blockstack/blockstack-browser/commit/8e4e70a3dbc62d9497b3b2c3cf5b33d25865f251

PR #1703, opened 2018-09-28
Fix social proofs verification in production web app

### LINKED ISSUE #1643: Proofs don't work unless you have installed blockstack downloadable
Proofs don't work on browser.blockstack.org because it tries to proxy them through the localhost CORS proxy that is only running if you've installed the windows, linux or mac browser distribution.

Thanks for @friedger for noticing this problem while working on the Android SDK.

craigiansmith:
This closes #1643. Previously social proofs were not showing as verified. As @yknl pointed out, this was because the app didn't recognise when it was running as a production build.

`isWebAppBuild()` now checks `NODE_ENV` and `WEBAPP` environment variables and returns true if `WEBAPP` is set and the `NODE_ENV` is "production".

The mac build target sets `NODE_ENV` to "production" aswell, but it doesn't set `WEBAPP` so it won't think it is a web app build.

This can be manually tested by running `npm run prod-webapp` then `npm start`. Logging into the browser at `localhost:5000` with a profile with a social proof that is verified, and going to `/profiles` to see the blue tick.

Another way of checking is to check the hostname. There is in fact code to do this in an earlier commit in this PR. However, I have reverted it due to the fact that `isWebAppBuild()` refers to the build process, not the running environment. In case we need another way to do it.

### DIFF
--- a/app/js/utils/window-utils.js
+++ b/app/js/utils/window-utils.js
@@ -14,9 +14,10 @@ export function isWebAppBuild() {
   const URL = document.createElement('a')
   URL.href = window.location.href
   const hostname = URL.hostname
-  return hostname === 'browser.blockstack.org'
-  //let isWebAppBuildCompileFlag = false
-  //return isWebAppBuildCompileFlag === true
+
+  const isWebAppHostnameFlag = hostname === 'browser.blockstack.org'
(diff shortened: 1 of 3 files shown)


craigiansmith:
@yknl could you review this PR?

yknl:
Thanks @craigiansmith I will take a look!

PR #1716, opened 2018-10-17
Fix image path in Registration Submitted View

friedger:
This PR
* fixes the broken path for the image in confirmation view after the user submitted a registration name.

Steps to see the broken path
1. submit a registration for a new username
1. See something like below
bildschirmfoto von 2018-10-17 08-54-36


yknl:
Thanks @friedger. This fixed the broken image for me.

Issue #1717, opened 2018-10-24
Users with zonefiles and non-standard profile paths (with no such profile written) cannot write to profile


prabhaav:
See: https://github.com/blockstack/blockstack-core/issues/875

The profile provided for Blockstack ID naval.id features a zonefile URL that cannot be loaded, blocking this user's access to GAIA storage. Go to https://core.blockstack.org/v1/names/naval.id and try to load the zonefile URL provided (https://gaia.blockstack.org/hub/19rJSaipwUdFwS5iD7yMvAvtAYGA36fTDS/0/profile.json). It returns a BlobNotFound.

The problem can also be seen when trying to use GAIA gateway, https://gaia-gateway.com/naval.id/www.stealthy.im/pk.txt, which returns: {"error":"Failed to load profile for naval.id"}.

The search endpoint won't even return a result for user naval.id: https://core.blockstack.org/v1/search?query=naval. It only returns one result for a different user "navaladi.id.blockstack".

We are seeing a variety of different behaviors for other important Blockstack user IDs as well, where the following behavior legend applies:

X - doesn't find
O - finds
D - downloads the profile

naval.id:
  - https://core.blockstack.org/v1/names/naval.id
  X https://gaia-gateway.com/naval.id
  X https://gaia.blockstack.org/hub/19rJSaipwUdFwS5iD7yMvAvtAYGA36fTDS/0/profile.json
  O https://gaia.blockstack.org/hub/19rJSaipwUdFwS5iD7yMvAvtAYGA36fTDS/profile.json

fred.id
  - https://core.blockstack.org/v1/names/fred.id
  O https://gaia-gateway.com/fred.id
  O https://gaia.blockstack.org/hub/1CER5u4QXuqffHjHKrU76iMCsqtJLM5VHu/0/profile.json
  X https://gaia.blockstack.org/hub/1CER5u4QXuqffHjHKrU76iMCsqtJLM5VHu/profile.json

fredwilson.id
  - https://core.blockstack.org/v1/names/fredwilson.id
  X https://gaia-gateway.com/fredwilson.id
  D https://gaia.blockstack.org/hub/1GSFLzRi4BzC4iQKDHypAfC5e4esYeM7DK/0/profile.json
  X https://gaia.blockstack.org/hub/1GSFLzRi4BzC4iQKDHypAfC5e4esYeM7DK/profile.json

werner.id
  - https://core.blockstack.org/v1/names/werner.id
[...]

PR #1718, opened 2018-10-24
Handle case where profile is missing for restored identity

kantai:
This fixes the underlying cause for the behavior experienced in #1717 -- the profile writing logic in the browser depends on the `identity.zonefile` entry being filled, which it wasn't in the case of profiles that fail to resolve (even though the zonefile resolves).

We may want to do something more than just this one-line fix though, such as:

1. pre-emptively writing the DEFAULT_PROFILE to the profileURL in the zone file location
2. calling refreshIdentities in the login flow if the identity.zoneFile field is null (this would allow such users to login correctly without having to go through the restore process again). Roughly around here: https://github.com/blockstack/blockstack-browser/blob/84ec3e261beaad57b844c37eb405f1649484d94a/app/js/auth/index.js#L218

### DIFF
--- a/app/js/profiles/store/identity/actions.js
+++ b/app/js/profiles/store/identity/actions.js
@@ -394,6 +394,7 @@ function refreshIdentities(
                         `refreshIdentities: resolveZoneFileToProfile for ${nameOwned} error`,
                         error
                       )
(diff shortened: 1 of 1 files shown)


yknl:
Thanks for the quick fix @kantai

PR #1723, opened 2018-11-05
Add developer-optional Gaia hub choice during onboarding

yknl:
This pull request is a proposed UX flow for user-selectable Gaia hubs. It does not actually change the Gaia hub yet.


moxiegirl:
The language on these strings is oriented to "Gaia" as opposed to the thing the user is setting. *Storage Provider* is the option they will be setting in browser settings. Gaia is our name for the underlying tech.

We should emphasize the wording that conveys what the user will encounter:

image

image

yknl:
I tested this using a locally running browser instance linked with the `solicit-gaia` branch of blockstack.js and a newly generated app also linked with the`solicit-gaia` branch. During the onboarding flow the browser correctly prompted for a custom Gaia hub URL. 

I first tested by entering the default Gaia hub URL `https://hub.blockstack.org`. The result was that writes to the hub succeeded. But there were some issues with the read URL.

`https://gaia.blockstack.org/hub//1AcE5iqQwxvrAkGc5UGddEvKKxUGXU1A9S/0/profile.json`
image



I ran a Gaia hub from the latest AWS AMI and entered the hub URL into the onboarding Gaia prompt. The result was that Gaia hub reads failed. Which most likely is because of the `read_url_prefix`. 
`{"challenge_text":"[\"gaiahub\",\"2018\",\"hub\",\"blockstack_storage_please_sign\"]","latest_auth_version":"v1","read_url_prefix":"hub"}`

image

yknl:
I tested this using a locally running browser instance linked with the `solicit-gaia` branch of blockstack.js and a newly generated app also linked with the`solicit-gaia` branch. During the onboarding flow the browser correctly prompted for a custom Gaia hub URL. 

I first tested by entering the default Gaia hub URL `https://hub.blockstack.org`. The result was that writes to the hub succeeded. But the read URL was constructed incorrectly. It has an extra `/` in the middle which caused reads to fail.

The incorrect profile.json read URL:
`https://gaia.blockstack.org/hub//1AcE5iqQwxvrAkGc5UGddEvKKxUGXU1A9S/0/profile.json`

What it should be:
`https://gaia.blockstack.org/hub/1AcE5iqQwxvrAkGc5UGddEvKKxUGXU1A9S/profile.json`

Network log for reference:
image


I also ran a Gaia hub from the latest AWS AMI and entered the hub URL into the onboarding Gaia prompt. The result was that Gaia hub reads failed. Which is most likely the same issue and that the  `read_url_prefix` might also be incorrect. 
`{"challenge_text":"[\"gaiahub\",\"2018\",\"hub\",\"blockstack_storage_please_sign\"]","latest_auth_version":"v1","read_url_prefix":"hub"}`

image

I also think that before we accept a custom Gaia hub URL during onboarding, we should check that it is a valid Gaia hub. Which we can probably do by fetching `<Hub URL>/hub_info`

hstove:
Yeah, that's an incorrect `read_url_prefix`. It needs to be a full URL (see https://hub.blockstack.org/hub_info).

> I also think that before we accept a custom Gaia hub URL during onboarding, we should check that it is a valid Gaia hub. Which we can probably do by fetching <Hub URL>/hub_info

We do fetch this information, but we don't have an error state if it fails.

yknl:
> Yeah, that's an incorrect read_url_prefix. It needs to be a full URL (see https://hub.blockstack.org/hub_info).

The `read_url_prefix` for the AWS AMI is not setup correctly, but I tried entering `https://hub.blockstack.org` which also resulted in incorrect read URLs. Were you able to setup a new account with `https://hub.blockstack.org`?

hstove:
I just checked it, and I see the error. The issue is that the `read_url_prefix` is `https://gaia.blockstack.org/hub/` (note the trailing slash). Then, it writes the profile to `https://gaia.blockstack.org/hub/asdf/profile.json`, but tries to fetch the profile from `/hub//asdf/profile.json`. It seems we have some mis-matched logic around fetching the profile.json with a slash in the front. I'll have to figure out some way around this, or add a standard that all `read_url_prefix` URLs must have a trailing slash.

PR #1729, opened 2018-11-24
Open new window when opening Github from Blockstack Browser

DougAnderson444:
Changed "Run your own Gaia storage hub" now opens a new browser window with the added `<target="_blank">`, so that the user's Blockstack account stays open in the browser


aulneau:
Nice! thanks for this, we should also add `rel="noopener noreferrer"`, see -> https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

hstove:
Thanks for this! I'm going to merge and add Thomas's suggestion.

PR #1755, opened 2018-12-20
Fix wallet API settings state bug

yknl:
Bump redux state version and only preserve API settings that should not change on update instead of everything.


kantai:
This looks good to me!

Issue #1759, opened 2018-12-22
Ensure user is routed back to browser instead of system default during authentication


aviaryan:
I just got an email from a user of [yournote.app](https://yournote.app) saying

> When I click on green sign in button nothing happens. Just presses then nothing. I am using latest Firefox 64 and even tried with adblocker turned off.

I tried it on my computer as well and since I use local Blockstack Browser, Sign in button opened the link in the default browser (which is Chrome) and then redirected it back to Chrome. So, trying to login in Firefox actually logged me in, in Chrome.

So, there are 2 issues here -
1. Blockstack sign in button doesn't work for someone who has Firefox as default browser
2. Sign in on Firefox works incorrectly for someone who is not using Firefox as the default browser

Thanks.

> [Cross-posted from this forum topic](https://forum.blockstack.org/t/issue-with-firefox-sign-in/7012)

jfacoustic:
Any updates on this?  I was able to reproduce it as well.

markmhx:
I believe there are two scenarios here:

1. If the desktop Firefox user does *not* have the Blockstack Browser installed locally (and therefore needs to use the hosted browser at browser.blockstack.org), Firefox fails to redirect the user to the hosted browser and simply does nothing (aka "When I click on green sign in button nothing happens"). This issue is being tracked with https://github.com/blockstack/blockstack.js/issues/558 and I believe the best interim solution on the app developer's end may be to detect the user agent and guide any desktop Firefox users to download the Blockstack Browser locally or use Chrome, etc. to access the app with the hosted browser.

2. If the desktop Firefox user *does* have the Blockstack Browser installed locally but it's *not* set as the computer's default browser, then they'll be directed into Chrome or whatever browser is the default browser to authenticate with it. The only way I know of preventing this browser-changing behavior is setting the default browser to Firefox (h/t @jehunter5811). I'll open an issue (or see if we have an existing one) to see if we can prevent this behavior regardless of the default browser setting.

markmhx:
Thanks for the report @alextibbles. I'm assigning this issue over to @zone117x since he's actively working on browser support at the moment. 

@zone117x You're already working on the first scenario I listed above over at https://github.com/blockstack/blockstack.js/issues/558, so I'm changing this issue's title to reflect the second scenario. 

Do you think there's anything we can do with this second scenario insofar as ensuring that the user is routed back to the browser they used to initiate auth, regardless of the system's default browser setting (e.g. with a parameter)?

[...]

PR #1774, opened 2019-01-04
Support for new protocol detection technique

zone117x:
Creating PR only to get a build from netlify.

### DIFF
--- a/app/js/auth/index.js
+++ b/app/js/auth/index.js
@@ -7,12 +7,14 @@ import { connect } from 'react-redux'
 import { AuthActions } from './store/auth'
 import { decodeToken } from 'jsontokens'
 import { parseZoneFile } from 'zone-file'
+import queryString from 'query-string'
 import {
   makeAuthResponse,
   getAuthRequestFromURL,
   redirectUserToApp,
   getAppBucketUrl,
-  isLaterVersion
+  isLaterVersion,
+  updateQueryStringParameter
 } from 'blockstack'
 import { AppsNode } from '@utils/account-utils'
 import {
@@ -115,6 +117,7 @@ class AuthPage extends React.Component {
     this.state = {
       currentIdentityIndex: this.props.defaultIdentity,
       authRequest: null,
+      echoRequestId: null,
       appManifest: null,
       coreSessionToken: null,
       decodedToken: null,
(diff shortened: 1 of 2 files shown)


hstove:
(Posting here so I remember)

URL to test the auth flow: https://zone117x.github.io/blockstack-protocol-ping-demo/public/

kantai:
@zone117x -- how would you recommend testing this? Does the following seem right?

1. Ensuring that the auth flow works normally for current applications.
   a. Creating new account
   b. With existing account
2. Ensuring that the auth flow works normally for the test app built with the prototype protocol detection.

Does that seem right?

zone117x:
@kantai Yes, I think both those tests should validate backwards compatibility.

jeffdomke:
Tested only in Chrome and Safari. Amazing. 👏👏👏

Safari: I'm noticing that when I have the browser installed, and click "open" it will open the local browser in Chrome (since it is default). Ideally, the behavior would be: Open an app in Chrome = open local browser in Chrome. Assume this is expected behavior but it caught my attention. 

Safari: Notice the little weird scrolling jumpiness.

Those are tiny issues, think we should ship even if they can't be fixed. Well done @zone117x

zone117x:
> Ideally, the behavior would be: Open an app in Chrome = open local browser in Chrome. 

This issue is pre-existing and being tracked here: https://github.com/blockstack/blockstack-browser/issues/1759#issuecomment-451953973

zone117x:
@hstove This PR currently does not perform the redirect ping if the local browser has no account logged in. Which causes the web app in the original tab to redirect to the hosted browser. 

This appears to be the area that would need modified https://github.com/blockstack/blockstack-browser/blob/3cfb20426ea52997a88f978b78e35786d54683d7/app/js/routes.js#L225
I'm not familiar with React and adding the redirection code here seems like a more risky change that would require additional review. 
Any idea of a quick and safe way to do fix this?

hstove:
> @hstove This PR currently does not perform the redirect ping if the local browser has no account logged in. Which causes the web app in the original tab to redirect to the hosted browser.

Fixed!

zone117x:
> > @hstove This PR currently does not perform the redirect ping if the local browser has no account logged in. Which causes the web app in the original tab to redirect to the hosted browser.
> 
> Fixed!

That was fast 👍

PR #1776, opened 2019-01-07
Hotfix to correct gaiaHubConfig info in profile.json

hstove:


hstove:
The way the state update logic works is that it updates the profile.json for all local identities you have in the browser. This should be sufficient, because the only effected `profile.json`'s are the ones you've logged into with in your browser.

It iterates and updates all profile.json's before even updating state.

When updating the profile.json, it creates a new Gaia token before signing, so any effected Gaia tokens in the profile.json will not be used. This means they can be invalidated before doing the state update, and the write will still work.

app/js/utils/profile-utils.js
```diff
@@ -175,7 +175,9 @@ export function signProfileForUpload(profile, keypair, api) {
     profile = {
```
kantai: Can we add an `if (profile.api) ...` so that it will remove the gaiaHubConfig entry if it exists even if `api` isn't passed in (this happens if you manually edit the profile, or the app sign in, I think)

kantai:
> When updating the profile.json, it creates a new Gaia token before signing, so any effected Gaia tokens in the profile.json will not be used. This means they can be invalidated before doing the state update, and the write will still work.

Yep -- can confirm that this is the case: the uploadProfile function here https://github.com/blockstack/blockstack-browser/blob/f2b12bf51123a696ef64fd65c730ebe050e04eb2/app/js/account/utils/index.js#L84
always tries to generate a new token on the fly, anyways.

app/js/utils/profile-utils.js
hstove: 👍 good call, updated

kantai:
This worked for me as well, @hstove

app/js/update/index.js
```diff
@@ -137,51 +140,85 @@ class UpdatePage extends React.Component {
...
-              this.props.updateState()
+    const updateProfileUrls = localIdentities.map((identity, index) =>
+      new Promise(async (resolve, reject) => {
+        try {
+          const signedProfileTokenData = signProfileForUpload(
+            identity.profile,
+            this.props.identityKeypairs[index],
+            this.props.api
+          )
+          console.log('newProfile', signedProfileTokenData)
```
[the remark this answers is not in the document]
hstove: :+1:

hstove:
Thanks for the reviews. I also removed the calls to add `browserVersion` in the gaia token payload, as it's no longer necessary.

Issue #1730, opened 2018-11-26
Next button with email does not move on in Microsoft Edge


moxiegirl:
Unable to move through the entire onboarding process in IE.  When I got to entering the email, I could press Next but the dialog just 

![image](https://user-images.githubusercontent.com/1347209/49023620-79519b80-f14c-11e8-9980-23339c32b655.png)

Version:

![image](https://user-images.githubusercontent.com/1347209/49025659-fb43c380-f150-11e8-92ce-7037abb3b3ea.png)

![image](https://user-images.githubusercontent.com/1347209/49025780-3940e780-f151-11e8-9ee7-3371e71deb5e.png)

markmhx:
@zone117x Has this issue's solution been completely reviewed and as such can be closed in anticipation of release?

Issue #1725, opened 2018-11-20
User enters email into username and no validation shows


jeffdomke:
![Screen Shot 2018-11-20 at 7:53 PM](http://marker.screenshots.prod.s3.amazonaws.com/17887e137bb2473ca1ec873fb6a4d39c-1542743584776.png)

Edge
32
Intel(R) Celeron(R) CPU N3350 @ 1.10GHz
3324.7 MB
1013.91 MB
Windows 10 Home 10.0
1366 x 768

3936.28 MB

---
### More Info

**Source URL**: [file:///Users/jeffdomke/Desktop/Screen%20Shot%202018-11-20%20at%202.51.20%20PM.png](file:///Users/jeffdomke/Desktop/Screen%20Shot%202018-11-20%20at%202.51.20%20PM.png)

markmhx:
@zone117x Has this issue's solution been completely reviewed and as such can be closed in anticipation of release?

Issue #1698, opened 2018-09-27
Edge Browser – cannot get any buttons to work


jeffdomke:
To test just use browserstack and latest edge and try to create an ID or restore.

aulneau:
I think we should punt this and do a larger audit of functionality across browsers.

zone117x:
So far only problem I'm seeing in Edge is a bug when trying to get past the "enter email" screen. 
I'm currently testing Edge support for the protocol handler detection so going to assign to myself. 
This issue also fits in an important KR for this quarter.

markmhx:
@zone117x Want to tackle this in the new sprint starting today?

zone117x:
Let's try to get this done this sprint, but debugging these Edge issues can be quite a time sink. So I'd punt this if it started looking like a multi-day debugging session.

PR #1793, opened 2019-01-15
Fix account creation and recovery in Edge browser

zone117x:
This fixes account creation and recovery when using Edge browser. 
The problem was Edge not supporting the use of crypto APIs within web workers. 
This PR simply does a runtime check for crypto API support in web workers, and falls back to doing the work on the main thread when not supported. 

Fixes 
https://github.com/blockstack/blockstack-browser/issues/1730
https://github.com/blockstack/blockstack-browser/issues/1725
https://github.com/blockstack/blockstack-browser/issues/1698

Does not fix: 
* Some css/positioning issues with placeholder text. 
* Bugs with Edge versions earlier than 44.


zone117x:
Added @aulneau as reviewer since you may have some input on the cross-browser supported related changes, and since Ken is on vacation.

zone117x:
Something to note - the exception causing this problem was never logged to console, and Edge developer tools were too cumbersome "break on all exceptions" and filter through the hundreds to find this one.
Identifying this bug required enabling trace logging, then putting try/catch/logs around suspicious areas of code.

Anyone with more React experience know how we can get these exceptions to be more prominent, as opposed to just having the GUI glitch out?

hstove:
When I test on Edge, I'm using the Netlify preview, and I can't even get past the step where I enter my desired username. I click continue and nothing happens. This might be related to #1698 

edge browser issue

zone117x:
Dang. I only tested the login to existing account flow. @hstove can you confirm that at least is fixed with this PR?
I'll look into account creation as well.

hstove:
> I only tested the login to existing account flow. @hstove can you confirm that at least is fixed with this PR?

Confirmed, works well.

aulneau:
Just tested this on Edge, works for recovery and new user registration!

zone117x:
@hstove I just tested account creation on Edge and it is working. Only problems I'm seeing are some css positioning issues, and the encryption/decryption process taking much longer in Edge than other browsers. 

Can you re-test?

yknl:
I just tested recovery using the deploy preview on Edge 42.17134 using a 12 word phrase. Clicking the "Sign In" button didn't do anything. The console logs also didn't show any errors.

zone117x:
@yknl Use Edge 44 (should just need to run Windows Update). There are significant problems with the DOM usage in earlier versions that I doubt we will ever fix. I think @hstove looked into some of them and could provide more details.

Issue #1790, opened 2019-01-14
Blockstack Mobile Browser Camera Bug


prabhaav:
When recovering an existing ID, if the user chooses to use the QR code flow, the following error comes up: 
<img width="768" alt="blockstackiosbug" src="https://user-images.githubusercontent.com/11723563/51132069-26a86d80-17e6-11e9-9022-415af1569155.png">

markmhx:
I believe we built this QR code feature only to decide to no longer support it. So we may be able to simply remove it. @jeffdomke Is that indeed the case?

I'm adding this to our current epic for ensuring that Blockstack works across systems and browsers, so we'll get to this ASAP.

markmhx:
@prabhaav We certainly want to unblock you here ASAP. @hstove Is there a quick fix we can get in place for Stealthy (perhaps by detecting user agents for web views like this?) that can resolve their app store need even if we need more time deciding on whether to keep QR codes in general?

markmhx:
Also @shreyasthiagaraj any ideas from our end on how we or they might resolve this with iOS code as a hotfix at least?

prabhaav:
This isn't in our iOS code, this is the Blockstack Auth that comes up before the user is even able to use the app.

markmhx:
@prabhaav Yea understood, just pinging him in case he has a creative solution given the web view context. But I imagine this will be something we resolve in the browser code.

hstove:
When I go to reproduce this, it asks me for permission to use the camera, and I click 'Accept'. Then it works as normal. If you click 'Deny', it shows this error message. Is that not your behavior?

prabhaav:
That question never comes up for me on mobile Safari

hstove:
You may have denied it previously. Go to Settings -> Safari -> Allow Camera & Microphone to see if it's denied

prabhaav:
This is a bug report filed by Apple app review team

hstove:
Yeah, I get that. So, it sounds like the only thing we could improve here is the error message? Unless there really is a bug here, but I wasn't able to reproduce it.

prabhaav:
![img_5b019bd55c2c-1](https://user-images.githubusercontent.com/11723563/51140908-9fb2bf80-17fc-11e9-95f0-f51c2843e35b.jpeg)

I have those settings enabled and I still see the same error

shreyasthiagaraj:
Yea, once the browser is open I don’t think we  have any control from iOS
[...]

PR #1794, opened 2019-01-15
Remove the ability to scan QR codes

hstove:
Apparently, scanning QR codes has never worked in native iOS apps, due to the way they're set up to use Safari. It's currently blocking Stealthy (discussed in #1790), so the easiest solution for now is to remove the button that lets you scan.

I believe we'd like to push this hotfix to the hosted browser, so that Stealthy can get past iOS approvers.

(cc @prabhaav for visibility)

QA process: view the netlify preview. Click 'sign in with an existing ID'. There should be no 'scan' button on the bottom of that modal. Everything else should work as normal.

### DIFF
--- a/app/js/sign-in/index.js
+++ b/app/js/sign-in/index.js
@@ -124,7 +124,7 @@ class SignIn extends React.Component {
       })
     })
 
-  backToSignUp = () => browserHistory.push({ pathname: '/sign-up' })
+  backToSignUp = () => browserHistory.push(`/sign-up${document.location.search}`)
 
   isKeyEncrypted = key =>
     import(/* webpackChunkName: 'bip39' */ 'bip39').then(
--- a/app/js/sign-in/views/_initial.js
+++ b/app/js/sign-in/views/_initial.js
@@ -89,7 +89,6 @@ export default class InitialSignInScreen extends React.PureComponent {
       })
     }
 
-
     const screen = {
       title: {
         children: 'Enter Secret Recovery Key or Magic Recovery Code'
--- a/app/js/sign-in/views/_initial.js
+++ b/app/js/sign-in/views/_initial.js
@@ -70,6 +70,25 @@ export default class InitialSignInScreen extends React.PureComponent {
   render() {
     const { next, value, ...rest } = this.props
     const { isScanning, scanError } = this.state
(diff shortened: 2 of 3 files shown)


moxiegirl:
@hstove if there are users that copied their original QR code but not the string...can they still use that code?

hstove:
@moxiegirl how does one copy a QR code?

kantai:
This worked for me -- but do we want to use the iOS auth flow param to toggle?

hstove:
I was able to successfully test and implement only hiding this in iOS. The iOS SDK adds `client=ios_secure` to the initial auth request. By passing this to the sign up view, we can easily check for it when rendering.

QA'ing this requires you to use Xcode and tweak the example app code slightly.

hstove:
> Tested by attempting to restore with and without the search code, and also tested signing up for a new name.

Nice. Yeah I didn't think of testing this just by hacking an auth request with an extra param.

moxiegirl:
@hstove I think we present it to them to take a picture when they first login.

hstove:
Thanks for the catch @shreyasthiagaraj . I've taken care of that and it works after you go back now.

shreyasthiagaraj:
@hstove The issue seems fixed on iOS.

**Minor nit:** The first time you go to "Sign in with an existing ID", the page to enter your recovery key loads and the keyboard is not up. But subsequent visits to this page after you go "back" result in the keyboard popping up with the recovery key textbox having focus. This isn't ideal because the text explaining what the recovery key and magic recovery code are is covered by the keyboard.

hstove:
> Minor nit: The first time you go to "Sign in with an existing ID", the page to enter your recovery key loads and the keyboard is not up. But subsequent visits to this page after you go "back" result in the keyboard popping up with the recovery key textbox having focus. This isn't ideal because the text explaining what the recovery key and magic recovery code are is covered by the keyboard.

Interesting. Probably worth tracking in a different ticket. We want autofocus on desktop but not on mobile.

hstove:
@kantai we're good to merge here I think, and then will you be able to kick off a web deployment from develop today?

The only other thing on develop that is merged is #1729, which is really small.

kantai:
Yep -- go ahead and merge to `develop` and then we can merge to the web deployment.

PR #1795, opened 2019-01-15
Support for building Windows app on Windows

zone117x:
Ported the shell build scripts and vsts config over to package.json scripts (similar setup to the macOS build scripts).

Building the Windows app and .msi now only requires running `npm run win32`

@kantai This is largely an additive PR, only couple minor changes to `BlockstackBrowser.csproj` and `BlockstackSetup.wixproj` to fix:
1. Long file paths when repo is not installed close to the drive root (ex: `C:\Users\matt\Documents\Source\Blah\Projects\blockstack-browser` vs `C:\blockstack-browser`
2. msbuild/batch path evaluation when deep git-bash integration into cmd.exe is not installed. 

Should only need to double check that the current msi producing CI works without issue.

PR #1801, opened 2019-01-16
Hotfix for webapp to disable QR on iOS

kantai:
Addresses #1790


kantai:
The changes in the hotfix are from PRs:

#1794, #1795, and #1729 -- each of those were individually tested, and the only consequential changes are in #1794, which were tested by Hank, myself, and Shreyas. Deploying to browser.blockstack.org now.

Issue #1760, opened 2018-12-30
Updating browser screen – add option to reset browser and start over with recovery key


jeffdomke:
![Screen Shot 2018-12-30 at 7:27 PM](http://marker.screenshots.prod.s3.amazonaws.com/bd61f88a9ac04bf883d74da1b33ad0aa-1546198035272.png)
---
### More Info

**Source URL**: [https://forum.blockstack.org/t/unable-to-log-in-after-browser-update/7079](https://forum.blockstack.org/t/unable-to-log-in-after-browser-update/7079)

markmhx:
@aulneau Could you provide some design guidance here when you have a chance?

aulneau:
Yep I will tomorrow!

jeffdomke:
Think we should prioritize this. The more stuff that breaks the more updates we push the more chances there are for this to frustrate users.

PR #1809, opened 2019-01-23
Allow reset on update page

aulneau:
### What this does

Whenever we update or change the structure of the application state object, we will increment the state version constant, forcing the end user to enter their password so we can restore their account after wiping everything. A common problem users have been running into is that there might be multiple passwords, or a user might just forget their password. Currently there is no easy way to reset the browser from this view. 

This PR adds the option to reset the browser from this screen, and then redirects the user to the sign in view.

I've also updated some of the dependencies in this project.

This resolves https://github.com/blockstack/blockstack-browser/issues/1760

### To do:
- [x] Ensure app auth requests persist through clearing of local data

screen shot 2019-01-23 at 10 35 06 am
screen shot 2019-01-23 at 10 34 57 am


aulneau:
Perfect, thank you @markmhx.

I don't think it makes sense to link `register a new Blockstack ID`, because the user still needs to do the act of resetting before being able to `sign-up`. I'll change the rest of it though!

aulneau:
screen shot 2019-01-23 at 1 40 58 pm

I changed 'remember' to 'have' in this sentence: `If you don't remember your Secret Recovery Key,` as I think it's more typical that people would store their key, and not ever need to 'remember' it.

I'm not sure what the button should say other than `reset` or `continue`. what do you think @markmhx?

markmhx:
@aulneau Looks good, maybe "Continue" best to keep it neutral and on the path to "sign in"?

aulneau:
@markmhx @hstove @zone117x Thanks for your reviews and feedback! I have merged the changes in :)

markmhx:
@aulneau Thanks for jumping on this! Excited to get it out.

PR #1810, opened 2019-01-23
Fixes `/clear-auth` endpoint

hstove:
When you clicked to confirm, it did actually reset your storage, but wasn't redirecting you properly. It made it seem like nothing happened.


hstove:
Yeah, @yknl would you mind giving this a quick review? I don't want to be breaking anything here by removing the `redirect_uri` param support, but I don't think it was working.

markmhx:
@aulneau needs to confirm that the redirect functionality works in web apps, and also that the intended functionality works on native mobile apps.

aulneau:
My PR fixes this, I think we can close this PR.

PR #1814, opened 2019-01-24
add SpringRole

kar2905:


hstove:
This is great. Thanks for the addition! We already cut out a beta release that we'll be pushing to production in a day or so, but this will be in the release after that. Sorry it took so long to merge!

PR #1803, opened 2019-01-16
Support for unsigned builds of the MacOS app

zone117x:
A few new build scripts for building the native MacOS app without having the $100 Apple developer certificate or needing access to the private PBC cert. 

Useful for CI automated building & `Blockstack.app` artifact publishing as well. See commit https://github.com/blockstack/blockstack-browser/commit/96c5c827130863ce751870675b85bf0dccefa2e3 for the travisCI build config, which we can PR after this one, if wanted.

PR #1815, opened 2019-01-24
Beta Mode

hstove:
This PR adds a new 'mode' to the MacOS app to enable "Beta Mode". In Beta mode, all auth requests are redirected to `https://develop--reporter-beaver-73821.netlify.com`. We'll probably want to change this URL to something like `beta.browser.blockstack.org` before releasing this feature.

This URL is auto-built for every new commit to `develop`. So, as soon as new things are merged to develop, you can test them without having to run a local environment.

This is great for internal testing, but it's also helpful for end users who run into breaking bugs. Instead of having their Browser be in a broken state until we push a release, we can at least let them know that they can try beta mode as soon as we have a fix (with understandable precautions).

Testing: alt-click the icon and then enable Beta Mode. When beta mode is enabled, you shouldn't be able to enable dev mode, and vice versa.

screenshot

This is built on top of #1803, because I needed that code to build and test a Mac app locally.


yknl:
My only issue with this feature is that when switching to beta mode, it's essentially signing the user out and asking them to restore again. Because they probably have never signed into `beta.browser.blockstack.org`. And it might cause confusion if they've signed into different ID's on `localhost` and `beta.browser.blockstack.org`. One thing we can do is automatically signing the user out when switching between beta mode and normal mode.

aulneau:
I wonder if there is a secure way to be able to transport the profile to the beta instance. We could send them to the link with the magic recovery code as a query, they'd just have to enter a password to restore. If they had previously logged in, the query param would just be ignored.

zone117x:
This is a cool feature. Should come in handy for internal usage and getting some % of the community using beta mode for us.

Agreed with @yknl comment. If we login for them, should we have them confirm that they are moving their potentially real blockstack keys from a "secure" local installation to a potentially less secure web hosted browser?

hstove:
Good feedback. OK - I'll update it so that, when you enable beta mode, you:

- See a page on localhost:8888 that says "do you want to migrate your account to a beta version of the browser?" - maybe some copy explaining this
- When you confirm, you are redirected to the beta URL with a param of the encrypted seed, so you can enter your password and it'll all be done.

hstove:
@markmhx @jeffdomke @aulneau I could use some design help here. You see this when you first enable Beta mode:

screenshot

What happens if you 'continue' is you get redirected to `beta.browser.blockstack.org`, with your encrypted seed as a URL parameter. You're prompted to enter your password to finish signing in to the beta version.

The copy is pretty janky. Any thoughts?

aulneau:
I think it's generally fine, but the first thing I notice is it's asking if you want to continue, but it provides no way to _not_ continue, eg a back or cancel button.

An alt proposal could be that when entered into beta mode, it can take you directly to the beta browser, with the enter password screen visible, and then that is the stage for them to determine if they want to continue or cancel.

markmhx:
If a user enables beta mode then disables it, will they retain their localhost session or does that get nuked upon the beta enable? I presume the former.

Here's how I'd recommend writing the copy to incorporate @zone117x's suggestion about hosted security as well:

> Ready to enable Beta Mode?
>
> By proceeding, your local Blockstack Browser will start redirecting all activity to a hosted version of the Browser at _beta.browser.blockstack.org_ that runs the latest code on [the develop branch](https://github.com/blockstack/blockstack-browser/tree/develop).
>
> Use it to test the latest code changes, but beware you'll have to authenticate your Blockstack ID again and you may encounter bugs. You also won't get any of the additional security benefits of using your local Browser until you disable beta mode.

And the buttons should be:

- `Yes, I'm ready` (purple –> enables beta mode and redirects to hosted beta homepage, with confirmation message on that page?)
- `No` (gray –> redirects to localhost homepage)

hstove:
The thing is, since this 'beta mode' feature uses native MacOS code, there isn't really a way to turn it off from within the app. You can only turn it off and on in the menu bar (like using Dev mode). 

So, I think the copy needs to be more like:

"You've enabled Beta Mode. From now on, Blockstack authentication requests will send you to beta.browser.blockstack.org. Would you like to securely log in using your current account?"

Continue -> go to beta and decrypt your key
No -> Beta is still enabled, but you won't be logged in

markmhx:
I see, so basically once they've chosen "Beta Mode" in the menu bar, it's enabled and the user just has to authenticate still.

How about this?

> You've enabled Beta Mode
>
> Your local Blockstack Browser will now start redirecting all activity to a hosted version of the Browser at _beta.browser.blockstack.org_ that runs the latest code on [the develop branch](https://github.com/blockstack/blockstack-browser/tree/develop).
>
> Use it to test the latest code changes, but beware you'll have to sign in with your Blockstack ID again below, and you may encounter bugs. You also won't get any of the additional security benefits of using your local Browser until you disable beta mode in the menu bar.

And the buttons should be the same as we have in the standard initial auth modal (for consistency):

screen shot 2019-02-28 at 11 46 37

hstove:
I agree, except I don't think we should say:

> You also won't get any of the additional security benefits of using your local Browser until you disable beta mode in the menu bar.

We don't indicate anywhere that the local build is 'more secure', and whether it is or not isn't crystal clear. It's mainly just "more decentralized".

markmhx:
Makes sense, how about just the following then for that paragraph?

>  Use it to test the latest code changes, but beware you'll have to sign in with your Blockstack ID again below, and you may encounter bugs. You can resume the usage of your local Browser at any time by disabling beta mode in the menu bar.

moxiegirl:
The tense "will now start redirecting" is very confusing. The beta *is* enabled it *is* redirecting -- present tense is now so you don't need the adverb and or the participle form.  Clarity.

> Your local Blockstack Browser is redirecting all activity to a hosted version of the Browser at _beta.browser.blockstack.org_ . The hosted version runs code on [the latest develop branch](https://github.com/blockstack/blockstack-browser/tree/develop). Be aware you may encounter bugs and you won't get any of the additional security benefits of using a local Browser until you disable beta mode in the menu bar.

> To disable Beta Mode, reset the browser and authenticate again.

markmhx:
"is redirecting" makes me feel like the currently viewed page is going to redirect automatically. i.e. present tense makes the redirection a bit _too_ immediate?

"will now redirect" maybe better? So that it has a sense of "going forward from now".

Separately per the last line in your copy suggestion, it appears the user won't have to authenticate again locally if they disable beta mode, though I may understand that wrong?

hstove:
> To disable beta mode, reset the browser and authenticate again.

No, it works by holding the option key and clicking the blockstack icon in the MacOS menu bar. Thats how you enable/disable it. You won't have to reset your browser when you turn it off/on.

moxiegirl:
> Your local Blockstack Browser is in Beta mode. All activity from now on redirects to a hosted version of the Browser at _beta.browser.blockstack.org_ . The hosted version runs code on [the latest develop branch](https://github.com/blockstack/blockstack-browser/tree/develop). Be aware you may encounter bugs and you won't get any of the additional security benefits of using a local Browser until you disable beta mode in the menu bar.

> To disable beta mode, on Mac, hold ⌥ OPTION KEY and click the Blockstack icon in your MacOS menu bar. In Windows, ...

@hstove is BETA mode only an option in Mac?

hstove:
With this pull request, it is. We would need to do additional work on other platforms to mimic this.

zone117x:
@moxiegirl 
> We would need to do additional work on other platforms to mimic this.

Would take like 15 min to add the same Windows context menu item - worries there.

zone117x:
Note to reviewers: this has been added to the Windows app as well
image

zone117x:
Will approve once we have https://github.com/blockstack/blockstack-browser/pull/1863 merged and deployed to `beta.browser.blockstack.org` for testing (unless we want to merge this PR first).

I have a UX request: after the redirect handoff it should only require entering the password. Its pretty annoying to have to go through the flow designed for on-boarding which is:
1) Get redirected to new URL which shows the `Save your Secret Recovery Key` form. Click `Secret Recovery Key`.
1) Enter password click `Next`.
1) Now on `Save all words` form, click `Continue`.
1) Now on `Select words #{x} and #{y}` form, go to 1password, click reveal, count the words, click the buttons (worst step). 
1) End up on `Save your Secret Recovery Key` form and click `Go to Blockstack`. 

We could add a query param (like `https://beta.browser.blockstack.org/seed?quick=1&encrypted=abc...`) and skip all the on-boarding forms/steps. 

Not a deal breaker at all for this PR, I can make a separate issue if necessary?

markmhx:
We decided during a review call to change the first button to "Create new ID or sign in" and remove the second one.

PR #1823, opened 2019-01-30
Update Apps Listing

aulneau:
### Staging
[Click for Staging](https://deploy-preview-1823--reporter-beaver-73821.netlify.com/)

### What this does
The apps in the browser have long been hardcoded, requiring a PR to update the order and apps in the list. This PR fetches data from the app.co api and displays the apps in two ways:

#### Top Apps
These are the top 15 apps from the most recent months ranking within the [App Mining](https://app.co/mining) program.

#### App Categories
The rest of the apps displayed are blockstack apps within their respective categories from within app.co. Some of these categories only have one app, often less than 10. I think a point of discussion is do we want to display these categories still? Also how do we want to order the categories?

--- 

Additionally, some of the apps still have broken images as a result of the SVG issue we recently fixed (cc @hstove)

The apps are display nearly the same as they were before, I have not opted to change the UI much at all for this PR.

screen shot 2019-02-21 at 9 56 00 am


jeffdomke:
This is dope. I do think there could be a good solution by listing in a similar manner to App.co. The names are often not that useful. And the icons aren't that important (could be minimized).

aulneau:
>  I do think there could be a good solution by listing in a similar manner to App.co. The names are often not that useful. And the icons aren't that important (could be minimized).

I agree, however I think if we opt to go for a redesign it will delay the release of this and also open a can of worms around the design of the rest of the application. (which I am fine having opened, but it felt like this could be an iterative thing).

aulneau:
I played around with displaying them in a similar way, but with 3-2-1 columns (responsive), without changing much of the look and feel. I'd be happy to take this opportunity to starting bringing the browser in line with our other interfaces (wallet, explorer).

screen shot 2019-01-30 at 4 53 49 pm

markmhx:
This looks generally great! Some notes:

---

Apps with long names should probably line-wrap instead of ellipse?

screen shot 2019-01-31 at 13 43 43

"Short" logos should probably be framed with top and bottom padding:

screen shot 2019-01-31 at 13 43 45

Apps without icon image values should perhaps show a default image or something other than the breaking image icon:
 
screen shot 2019-01-31 at 13 43 49

I think it's okay to show categories with one or few apps, especially if we show descriptions since that helps fill out the space aesthetically.

I was going to suggest we link to the App.co app profiles before seeing the version with app descriptions in the browser, but I think now that I've seen that, I like linking directly to the apps themselves. Question may come in the future how we want to link if apps have multiple destinations (e.g. web app, iOS app, etc).

I like the mockup of our new branding applied, though I'd suggest splitting off its application into a separate issue – and probably a broader one related to updating the browser's aesthetic entirely – so we don't get hung up on a redesign as part of the current sprint.

aulneau:
> "Short" logos should probably be framed with top and bottom padding

> Apps without icon image values should perhaps show a default image or something other than the breaking image icon:

Both of these are issues we should fix on the backend, eg replace non-square images with square versions, and the broken images are related to a different bug that @hstove has (almost?) fixed.

I agree with the rest of your comments, although I'm not sure about top 50! that's so many and there would be so many duplicates in that category along with the rest of the listing

markmhx:
@aulneau Got a new staging link? The one in the description (https://5c51d816b28770017c52dcca--reporter-beaver-73821.netlify.com/) appears to show the older design iteration:

screen shot 2019-02-12 at 16 22 37

markmhx:
It appears we need to line break single words that overflow in app descriptions (and maybe names as well?):

screen shot 2019-02-12 at 16 26 08

markmhx:
Do we want to show the light blue background color for logos instead of omitting it / letting it show through as white?

screen shot 2019-02-12 at 16 43 02
screen shot 2019-02-12 at 16 43 06

markmhx:
Can we conditionally not try to show images when the URL is missing?

screen shot 2019-02-12 at 16 45 01

markmhx:
What do we think of maybe putting these on one-per-line for mobile instead of two? Maybe it'd help readability? Two isn't bad but seems perhaps harder to scan in this context since reading the descriptions is so important (vs. a mobile home screen where you're already familiar with the apps' purposes).

screen shot 2019-02-12 at 16 46 45

markmhx:
Grammatical nitpik: "Top-ranked apps" here per [what I'm seeing with a quick google](https://english.stackexchange.com/questions/308815/hyphens-in-nationally-top-ranked).

screen shot 2019-02-12 at 16 49 01

markmhx:
Did we decide to order category sections (i.e. all sections other than the "Top-rated apps" one) by alphabetical order? That would seem a bit easier to scan for apps by name. 

screen shot 2019-02-12 at 16 50 28

markmhx:
Otherwise, looking sharp! Review complete.

markmhx:
Wait – one more thought – should we also order all category sections by alphabetical order? Or by number of apps? They seem quite random at the moment. I'd be happy with either strategy. Perhaps "total apps desc, alphabetical asc"

aulneau:
Perfect, thank you for your comments @markmhx, will implement these changes today.

aulneau:
@markmhx all changes have been implemented and should be reflected in stage

markmhx:
@aulneau want me to review the design implementation again or confident it's all good?

jeffdomke:
This looks great @aulneau. Two small things: 
- shadows look a bit funky, would reduce the blur or spread a bit. 
- I'm not loving the way, depending on the line break, some icons are above others. 

screen shot 2019-02-14 at 1 23 03 pm

pstan26:
So excited for this change, so many app founders are looking forward to this as well! Thank you in advance. :-)

markmhx:
This is a minor detail, so cool with me if we ship as currently designed, but I think my eye would prefer the columns to be consistent across rows like a table, with left alignment throughout:

screen shot 2019-02-22 at 18 45 41
screen shot 2019-02-22 at 18 45 36
screen shot 2019-02-22 at 18 45 32

markmhx:
We sure we want to call these "Popular Apps" since that could mistakenly indicate they get the most traffic / have the most users or votes? Rather than having been ranked highly e.g. "Top Apps", "Top-Ranked Apps"

screen shot 2019-02-22 at 18 47 56

markmhx:
Final nitpick, I might add a bit more padding between the logos and text.

But other than that, this looks great! Seems ready to 🚢 to me.

screen shot 2019-02-22 at 18 49 11

aulneau:
I am going to merge this once the tests pass. I changed `Popular` to `Top` per @markmhx's request.

PR #1824, opened 2019-02-04
[WIP] QA process document

hstove:


hstove:
Hey @timstackblock , in early February I took a shot at a "QA process document" for the Blockstack Browser. The idea is to lay out a rough process for peers to follow when doing code reviews, and also for beta testers to potentially evaluate beta releases that we put out.

As you can see, what I've put together is quite vague and I don't think it's very helpful. Now that you're on the team, I am curious if we should scrap this doc, iterate on it, or wait until we've decided on a more full-fleshed QA process that I know you're thinking about.

timstackblock:
@hstove thanks for sharing this, I do not think we should scrap it lets use it as a basis to build out a new document. We can consider what is missing and improve the process.

PR #1830, opened 2019-02-11
Feature/private gaia hub support

jcnelson:
This PR adds support for private Gaia hubs for applications, where the user's application Gaia hub is _distinct_ from the profile Gaia hub.  This is necessary to allow _existing_ Blockstack IDs to be used with _new_ Gaia hubs, without having to alter the zone file.

With this PR, it becomes possible to store your profile on hub A, and your app data on hub B.  For example, my blockstack ID `judecnelson.id`'s profile is stored on `https://gaia.blockstack.org`, but my app data is stored on `https://blockstack-server-profiles.s3.amazonaws.com`.  This PR adds a new API setting called the "Profile Gaia Hub Connection," and renames the existing Gaia hub API setting to "Application Gaia Hub Connection."  The former would be `https://hub.blockstack.org` (in my case), and the latter would be `https://my-private-gaia-hub`.  In addition, this PR makes it so the `api` section of my profile will store both the write URLs for the app Gaia hub as well as my profile Gaia hub.

### Signing In 

* Set up a private Gaia hub and whitelist your ID-address
* Set the private Gaia hub's write URL as your "Application Gaia Hub Connection" in your API settings
* Make sure your "Profile Gaia Hub Connection" URL is still `https://hub.blockstack.org`
* Sign into a Blockstack app
* Confirm that you can store data to your private Gaia hub
* Confirm that in your profile, your `apps` entry for the app points to your private Gaia hub's read URL
* Confirm that your latest profile is stored to your "Profile Gaia Hub Connection" (i.e. _not_ your private Gaia hub, but the Gaia hub pointed to in your zone file)
* Confirm that in your profile, the `api` entry contains the following:
   * The `gaiaHubUrl` value should be set to your private Gaia hub's write URL
   * The `profileGaiaHubUrl` value should be set to your profile Gaia hub's write URL (i.e. `https://hub.blockstack.org`)

### Recovery

* Go through the Blockstack ID recovery process on a profile that you signed into with this PR (i.e. so you have the `profileGaiaHubUrl` field set)
* Confirm that your private Gaia hub URL and profile Gaia hub URL values are restored in the API settings.


jcnelson:
I'll go ahead and split this into two PRs -- one that adds association tokens, and one that adds a setting for switching your app data Gaia hub.

jcnelson:
Superceded by https://github.com/blockstack/blockstack-browser/pull/1833

PR #1831, opened 2019-02-12
MacOS: Route user back to browser instead of system default during authentication

### LINKED ISSUE #1759: Ensure user is routed back to browser instead of system default during authentication

zone117x:
Fixes #1759 for MacOS.

Tested with Chrome, Safari, Firefox, Opera, and Brave. 

Apple provides a pretty straightforward API to accomplish this, nevertheless, this PR includes a bunch of defensive error handling in case some untested web browser refuses the open URL request, or cannot be mapped to a MacOS application via the originating event info.

### DIFF
--- a/native/macos/Blockstack/Blockstack/AppDelegate.swift
+++ b/native/macos/Blockstack/Blockstack/AppDelegate.swift
@@ -48,6 +48,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
     let log = OSLog(subsystem: Bundle.main.bundleIdentifier!, category: "Default")
 
+    func applicationWillFinishLaunching(_ notification: Notification) {
+        // This must be registered before `applicationDidFinishLaunching` otherwise the event
+        // may not be triggered if Blockstack.app was not already running.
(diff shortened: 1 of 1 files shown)


zone117x:
Thanks @hstove! Spent a couple years building a product using https://github.com/xamarin/xamarin-macios so vaguely familiar with a lot of the apple APIs.

Issue #200, opened 2017-02-09
npm install fail on web-0.3 branch


cstrangedk:
This is still happening on the `web-0.3` branch.  Ref: #199 and #125 

**Steps to reproduce:**
`git clone https://github.com/blockstack/blockstack-browser.git -b web-0.3`
`npm install`

**Error:**

`npm ERR! git clone --template=~/.npm/_git-remotes/_templates --mirror ssh://git@github.com/shea256/ecurve.git ~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494: Cloning into bare repository '~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494'...
npm ERR! git clone --template=~/.npm/_git-remotes/_templates --mirror ssh://git@github.com/shea256/ecurve.git ~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494: Permission denied (publickey).
npm ERR! git clone --template=~/.npm/_git-remotes/_templates --mirror ssh://git@github.com/shea256/ecurve.git ~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494: fatal: Could not read from remote repository.
npm ERR! git clone --template=~/.npm/_git-remotes/_templates --mirror ssh://git@github.com/shea256/ecurve.git ~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494:
npm ERR! git clone --template=~/.npm/_git-remotes/_templates --mirror ssh://git@github.com/shea256/ecurve.git ~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494: Please make sure you have the correct access rights
npm ERR! git clone --template=~/.npm/_git-remotes/_templates --mirror ssh://git@github.com/shea256/ecurve.git ~/.npm/_git-remotes/git-ssh-git-github-com-shea256-ecurve-git-cd23c494: and the repository exists.
npm ERR! git fetch -a origin (ssh://git@github.com/shea256/bitcoinjs-lib.git) Permission denied (publickey).
npm ERR! git fetch -a origin (ssh://git@github.com/shea256/bitcoinjs-lib.git) fatal: Could not read from remote repository.
npm ERR! git fetch -a origin (ssh://git@github.com/shea256/bitcoinjs-lib.git)
[...]

PR #1833, opened 2019-02-13
Feature/gaia association token

jcnelson:
This PR adds support for generating and passing along a Gaia association token on sign-in.  This will allow users who onboarded with a private Gaia hub to use that Gaia hub without having to whitelist individual app addresses.

## How To Test

* Start up your own Gaia hub, but do _not_ whitelist any addresses yet.  This is necessary for onboarding to work.
* Make a new Blockstack ID and onboard with your custom Gaia hub.
* Edit your Gaia hub's config to whitelist just your ID-address, and restart it.
* Proceed to sign in to an app with your new Blockstack ID.
* Confirm that you are able to use apps normally -- your Gaia hub should be receiving and storing data, and the auth header should contain an association token.
* Confirm that your Blockstack ID's zone file points to your custom Gaia hub, and that your profile's `apps` point to your custom Gaia hub.

You can see the results in my test Blockstack ID `testlocalgaiahub.test-personal.id`.


hstove:
This looks good to me. I'm going to use this in a sample app to test it out.

markmhx:
@kantai Were you able to review this change, and does it look good to you?

markmhx:
@jcnelson Also, any remaining changes you want to make here or good for QA then release?

app/js/auth/index.js
```diff
@@ -407,6 +431,13 @@ class AuthPage extends React.Component {
     if (appRequestSupportsDirectHub(this.state.decodedToken.payload)) {
       hubUrl = this.props.api.gaiaHubUrl
     }
+    if (isLaterVersion(requestVersion, '1.3.0')) {
```
kantai: Is version 1.3.0 necessary? Since the association token is just an additional field in the auth response, it should be okay to include regardless.

jcnelson: Yeah, that check can be removed.  I put it in there for clarity and sanitation -- a client that doesn't know about association tokens shouldn't receive them (but it won't break them if they do).

markmhx:
@kantai @hstove @yknl Please review during the sprint starting today

moxiegirl:
@markmhx Won't we want to update the Gaia user documentation with this?  Can I get  user facing use case for this  --- I've only been following this a bit.

markmhx:
Good question, @moxiegirl. I'm not really sure. @jcnelson should we sync up this week to discuss documentation needs here?

jcnelson:
> On second thought - this doesn't add the association token to profile update requests. I think that's going to be necessary before merging this - otherwise you won't be able to update your profile.json if you use whitelisting.

@hstove can you clarify what you mean here?  What's a "profile update request" and why does an association token need to be added to one?

hstove:
> @hstove can you clarify what you mean here? What's a "profile update request" and why does an association token need to be added to one?

I think I my logic might have been wrong when making that comment. I was thinking you'd need to pass an association token when updating your `profile.json` file, but since that address is already whitelisted, you don't need an association token.

hstove:
I've done more extensive testing of this with my own Gaia hub and various apps. It works really well. I [found and fixed](https://github.com/blockstack/blockstack.js/pull/645) a bug in blockstack.js v19 that causes association tokens not to work, which is unfortunate.

Overall this is awesome, and now I'm all set to use my own Gaia hub with a new name I have.

moxiegirl:
@jcnelson @markmhx didn't get a response to my question about the docs.  Also, the current JSON schema I have for the hub configuration doesn't show the whitelist.

hstove:
@moxiegirl it would make sense to add the instructions for having a private gaia hub. I would be happy to help you with that, including the `config.json` attributes required to whitelist addresses.

However, it's only really relevant for the people running their own Gaia hub, but I'm not sure if that's what you meant by "end user".

moxiegirl:
Thanks for the response @hstove. By user, I mean someone running a Gaia hub. We have instructions for someone running their own Gaia hub already. We even discuss whitelisting in the admin section. So, perhaps we have enough...it appears it is necessary for the user to understand "association token" concept and how it looks in a header -- if only to confirm it is there.

hstove:
Hey @moxiegirl , I created a new issue in docs.blockstack for continuing this discussion, so that we can keep this PR focused on code review and QA.

https://github.com/blockstack/docs.blockstack/issues/198

moxiegirl:
Issue moved to [blockstack/docs.blockstack #200](https://github.com/blockstack/docs.blockstack/issues/200) via [**ZenHub**](https://www.zenhub.com/)

moxiegirl:
oops. sorry was moving the issue

markmhx:
@yknl is spinning up a hub to test this and @zone117x is also testing with his DO hub

app/js/auth/index.js
kantai: Okay, that's fine.

zone117x:
Ping @timstackblock for QA. 

It's a bit of work to get everything setup to test the private hub feature this provides. Just to recap and expand on Jude's instructions, you need to:
1. Setup a personal gaiahub server.
1. Register a new name with the zonefile pointed at your personal gaiahub. This requires the CLI or using an app like Graphite that lets you specify a custom hub.
1. The name must be a primary account, N+1 accounts will not work due to other bugs.
1. Setup a sample app like hello-blockstack, and have it using the develop branch of blockstack.js (`npm i blockstack@blockstack/blockstack.js#develop`). 
1. Perform a `putFile` and `getFile` and ensure the private hub is being used and working (via the web browser network dev tools or similar). 

I think since several of us already confirmed that worked, it may be adequate QA for now to just ensure there were no regressions with normal usage of the browser.

PR #1834, opened 2019-02-14
E2E test for sign-in-with-blockstack using hello-blockstack app

zone117x:
Includes:
* A barebones Hello-Blockstack app which gets self-hosted for this test. 
* Additional code for reducing flaky test failures on BrowserStack's grid. 
* Debug logs are now collected on test failures and added as CircleCI artifacts. 

Notes: 

* This new test suite can be ran locally, but if the native app is installed, then this test will be skipped for Chrome browser. This is due to limitations with Selenium being incapable of closing Chrome's "Open App?" dialog, and Chrome's recent change that prevents protocols from being ignored. See the [related comments in the code](https://github.com/blockstack/blockstack-browser/blob/2989d6ef23e1867db65bb1e91a235f985077b0b9/test-e2e/utils/create-test-suites.js#L250) for more details.
      EDIT: Due to Chrome being the only browser that supports the debug log API, a workaround for this has been implemented for macOS where the system protocol handler is temporarily unregistered when the test is ran. The macOS Launch Services database is ephemeral and gets restored on its own, or simply re-open Blockstack.app to register the handler again. 

* For the future devops setup which runs these test against production servers, we will need the helloblockstack.com site updated and pointed at, since the production endpoints will not be able to reach the locally hosted app manifest.json during auth validation. 

* This PR includes an minimum viable script for running the test suites in parallel to speed up execution again BrowserStack's grid. It's not being used for CircleCI (yet) - it needs more status reporting, and other approaches need explored since Mocha was not designed to run tests in parallel and becomes very brittle when forced to.


test-e2e/auth/sign-in.js
```diff
@@ -0,0 +1,96 @@
...
+  step('wait for auth page to load', async () => {
+    await driver.el(By.xpath('//div[contains(.,"Select an ID")]'));
+  });
+  
+  step('click allow auth button', async () => {
+    await driver.click(By.xpath('//span[text()="test_e2e_recovery"]'));
+  });
+
+  step('ensure logged into hello-blockstack app', async () => {
+    await driver.el(By.xpath('//div[contains(.,"Hello, Alice")]'));
```
kantai: Can we do some simple sanity checks on the authenticated session data? Like that the logged in user has an appPrivateKey, hubUrl, etc. ?

test-e2e/auth/sign-in.js
```diff
@@ -0,0 +1,156 @@
...
+      window.BLOCKSTACK_HOST = '${browserHostUrl}/auth';
+    `);
+  });
+
+  step('click login button', async () => {
+    const windowHandle = await driver.getWindowHandle();
+    await driver.click(By.css('#signin-button'));
+    await driver.sleep(1500);
+    if (await driver.elementExists(By.css('#signin-button'))) {
+      // This closes the "Open app?" dialog on Edge.
```
hstove: Dude you're insane with the workarounds 🥇

Issue #1167, opened 2018-01-14
"Sign In With Blockstack" wants to launch Office 2016


sondreb:
The links or mime-type of the "Sign In With Blockstack" is registered to Office 2016 on Windows, at least when running from Google Chrome.

I still works to click accept in the dialog, it will redirect properly in a new tab, but I'm sure it will scare away some users (including myself) from clicking it.

![image](https://user-images.githubusercontent.com/309938/34915022-76ee5670-f91e-11e7-9892-df1f23df6e42.png)

kantai:
This is probably related to the way we register the protocol handler in Windows --

What version of blockstack is installed on your computer?

sondreb:
I have insider build of Windows 10, so the very latest, and Office 2016. I have the latest version of blockstack (I attempted this right after installation and setup). If you need exact version, I'll give it later as I'm not at that computer right now.

demarant:
![graphite-sign-in-with-openstack-ubuntu17](https://user-images.githubusercontent.com/320440/37020579-8343f6b6-211c-11e8-9e32-382e94bff0a5.png)

I get a similar dialog box on on ubuntu 17.10. I am unable to sign in, did I miss anything?

mingrui:
same problem on windows.

@demarant same problem on ubuntu.

ddyson:
Is anyone working on this? I have just hit this problem half way through the Hello BlockStack tutorial. https://blockstack.org/tutorials/hello-blockstack

adsri:
Any update on this issue? It is still coming for windows10.

bronoman:
Issue seems to be tied to Chrome on Windows 10 with MS Office installed. No issues with [Brave](https://brave.com/).

hcollin8:
Issue is not resolved, any updates?

moxiegirl:
@yknl what do you think? I can test it if you are not sure.

yknl:
This issue needs further investigation. Seems to be a problem that affects both Windows and Linux. We'll move this up the priority list for the browser team.

zone117x:
No. But I know what's causing that on Windows and it's pretty easy to fix, maybe a few hours.

zone117x:
This PR https://github.com/blockstack/blockstack-browser/pull/1796 actually has some of the changes needed to fix it

markmhx:
@zone117x Want to tackle this in the new sprint starting today?

zone117x:
[...]

PR #1843, opened 2019-02-21
Window app protocol handler fixes

zone117x:
* Fixes https://github.com/blockstack/blockstack-browser/issues/1167 `"Sign In With Blockstack" wants to launch Office 2016`
* Fixes bug with auth not working when the local app is not running - which it is not running by default after installation or after machine restart. 
* Fixes bug with multiple app instances being openable but silently failing to start local webserver(s).
* Fixes Windows Firewall prompt/warning on first launch due to `blockstackProxy.js` listening on public network interface. 

To test the `launch Office` bug then install Microsoft Office 2016. I followed these instructions https://support.office.com/en-us/article/use-the-office-offline-installer-f0a85fe7-118f-41cb-a791-d59cef96ad1c#OfficePlans=signinorgid with the option `Install the 32-bit version of Office Business` 

Or you can run `regedit` and verify nothing like this registry entry exists `Computer\HKEY_CLASSES_ROOT\Local Settings\Software\Microsoft\Windows\Shell\MuiCache` > `..Blockstack\ProtocolHandler.exe` == `{something bogus}`.


zone117x:
👍 
@kantai did you happen to test out this PR at the same time? https://github.com/blockstack/blockstack-browser/pull/1844

kantai:
I tested them separately.

PR #1846, opened 2019-02-22
Tests for validating backwards compatibility in blockstack.js v19

zone117x:
Tests for more blockstack.js APIs for ensuring https://github.com/blockstack/blockstack.js/pull/542 remains backwards compatible. 

(Note: this is not an exhaustive list of tests for all public blockstack.js methods & parameters)

All tests are currently passing when running against the v19 PR when running 
```
npm install blockstack@blockstack/blockstack.js#feature/storage-strategies
TEST_E2E_GREP=login-to-hello-blockstack-app npm run test-e2e:browserstack
```

### DIFF
--- a/test-e2e/auth/sign-in.js
+++ b/test-e2e/auth/sign-in.js
@@ -71,8 +71,6 @@ createTestSuites('login-to-hello-blockstack-app', ({driver, browserHostUrl, loop
       await driver.switchTo().window(windowHandle);
       await driver.sleep(4000);
     }
-    // Failed attempt to close protocol handler prompt on chrome:
-    // await driver.actions({ bridge: true }).sendKeys(Key.ESCAPE).perform();
   });
 
   step('wait for auth page to load', async () => {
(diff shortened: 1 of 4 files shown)


hstove:
Are you sure this is running against v19? I say this because blockstack.js usually doesn't work with a git import from npm, because we don't commit the compiled version of the library to Github.

test-e2e/auth/sign-in.js
```diff
@@ -100,42 +97,80 @@ createTestSuites('login-to-hello-blockstack-app', ({driver, browserHostUrl, loop
...
 
-    // Verify file contents.
-    expect(getFileResult).to.equal(gaiaFileData);
+  step('validate blockstack.signUserOut(...)', async () => {
+    const redirectUrl = `http://${loopbackHost}:${helloServerPort}/?some=param`;
+    await driver.executeScript(`blockstack.signUserOut(arguments[0])`, redirectUrl);
+    await driver.sleep(50);
+    await driver.el(By.css('#signin-button'));
+    const windowLocation = await driver.getCurrentUrl();
+    expect(windowLocation).to.equal(redirectUrl);
```
hstove: Lets add a line to this test to make sure `localStorage.blockstack` doesn't exist after sign out

zone117x:
@hstove 
> Are you sure this is running against v19? I say this because blockstack.js usually doesn't work with a git import from npm, because we don't commit the compiled version of the library to Github.

It works for me (`npm i blockstack@blockstack/blockstack.js#feature/storage-strategies`). The webpack rebuild succeeds, and the self-hosted hello-blockstack app uses the `blockstack.js/dist/blockstack.js` file that we commit, using something like:
```js
const blockstackDistPath = require.resolve('blockstack/dist/blockstack');
app.use(express.static(path.dirname(blockstackDistPath)));
```

I also just ran with `npm i blockstack@19.0.0-beta.1` to confirm - all good.

test-e2e/auth/sign-in.js
zone117x: @hstove This has been implemented

hstove:
The account creation tests are failing, and I retried it after it didn't pass the first time. Mind fixing that before we merge?

zone117x:
@hstove Looks like the account-creation test is just extra flaky today. Its failing on a different browser environment on each run. That test wasn't changed with this PR, I think we are okay to merge.
It would take some time to diagnose the cause of flake and come up with some more anti-flake techniques.

Issue #1799, opened 2019-01-15
Improve UX of authentication clearance page


shreyasthiagaraj:
https://browser.blockstack.org/clear-auth

Clicking "Press again to confirm" at the Sign Out screen does nothing on web. On iOS I get this error: "Safari cannot open the page because that address is invalid."

markmhx:
I can confirm on my end, too, that it's not working. Has this broken recently? It seems it was working just a few days ago.

markmhx:
- Add /reset path
- Ensure redirect to /sign-up across all devices upon "Reset My Device" click
- Snazzy up design with standard components

markmhx:
@aulneau Does this issue still seem feasible for the current sprint?

Issue #1624, opened 2018-08-21
Non-square avatars get squished


wbobeirne:
We currently allow uploading avatars that are non-square, but they get rendered squished. We should do one of the following:

1) Resize the image by cropping the sides / letting the user crop when they upload.
2) Render all avatars as centered background images instead, and note that aspect ratio is not guaranteed in our documentation.

PR #1859, opened 2019-03-06
/sign-out path and misc ui updates

aulneau:
### What this does

This fixes a few issues:
- #1624 (square aspect ratio of avatars)
    - https://github.com/blockstack/blockstack-browser/pull/1859/commits/9b871f7b7b379d4cae5a3e6f33f37b4595f7cc14
- #1799 (`/clear-auth` -> `/sign-out`) 
    - https://github.com/blockstack/blockstack-browser/pull/1859/commits/00520f9b8851394c1543424ddf5da9d79c1aeb7a
- Refactored nav bar to use `mdi-react`

screen shot 2019-03-06 at 3 04 44 pm

screen shot 2019-03-06 at 2 58 15 pm


markmhx:
@aulneau Is the above unit test something you can fix soon so we can get this merged?

aulneau:
> @aulneau Is the above unit test something you can fix soon so we can get this merged?
I reached out to the group a couple weeks ago and most people agreed that many of these unit tests aren't really useful/are legacy and we don't need them. I will check in with @timstackblock and get back to you.

timstackblock:
> > @aulneau Is the above unit test something you can fix soon so we can get this merged?
> 
> I reached out to the group a couple weeks ago and most people agreed that many of these unit tests aren't really useful/are legacy and we don't need them. I will check in with @timstackblock and get back to you.

I have worked with the e2e test I think that these are valuable and should NOT be deleted. They have some flakes and should be refactored. There is an open issue for this and I am looking into it. 

As for unit test, my experience is that these are just developer test, and their primary purpose is to verify small components of the code during development. If the code is being refactored usually the developers would refactor the unit test at same time. 

I would defer to the developer who created them, they don't really test end to end, its possible for the unit test to pass and the QA engineer's e2e test to still fail because of integration issues so they are for development purposes.

aulneau:
Will resolve conflicts and try to get tests passing this week.

timstackblock:
Tested this still one small issue circle avatars

markmhx:
We've decided to address https://github.com/blockstack/blockstack-browser/issues/1624 in a separate PR.

PR #1863, opened 2019-03-09
Fix for seed recovery from email

hstove:
This initially came up in the 'beta mode' PR, but it turns out this is broken on production right now. We send out a recovery email when you sign up, with your encrypted seed phrase. There is a URL to restore that phrase, however, that flow actually doesn't work. It decrypts your phrase, but it never actually restores your account. Maybe that's the purpose of it, but I think you should get logged in. Especially because the new 'Beta Mode' uses this flow to log you in on the beta URL.

### DIFF
--- a/app/js/seed/index.js
+++ b/app/js/seed/index.js
@@ -7,15 +7,23 @@ import { connect } from 'react-redux'
 import { bindActionCreators } from 'redux'
 import { AccountActions } from '../account/store/account'
 import { IdentityActions } from '../profiles/store/identity'
+import { SettingsActions } from '../account/store/settings'
 import { ShellParent, AppHomeWrapper } from '@blockstack/ui'
 import {
   selectEncryptedBackupPhrase,
-  selectRecoveryCodeVerified
+  selectRecoveryCodeVerified,
+  selectIdentityAddresses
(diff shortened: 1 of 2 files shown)


zone117x:
For reference - this unblocks a bug identified in the Beta mode PR: https://github.com/blockstack/blockstack-browser/pull/1815#pullrequestreview-212435672

zone117x:
Created a QA issue for adding this recovery flow to the e2e test suites https://github.com/blockstack/blockstack-browser/issues/1864

yknl:
The original purpose of the `/seed` route was to allow retrieval of the seed phrase from an encrypted recovery code without affecting any existing sessions. 

If we're using this page to enable beta mode, the copy here doesn't make any sense. 

image

Additionally, in the recovery email sent after onboarding, the copy suggests that the link only let's you view your seed phrase. If going through the flow changes the user session, it might cause a lot of confusion.

image


I would suggest using a different page to restore a user session from encrypted recovery code when switching to beta mode.

markmhx:
We decided during review to close this PR out in favor of a different UX approach for https://github.com/blockstack/blockstack-browser/pull/1815

PR #1868, opened 2019-03-12
[WIP] New onboarding prototype

hstove:


moxiegirl:
@hstove Is this something where we are actually changing the flow? There is work involved in documentation, so it would be helpful to see the proposal so I can scope my work.

hstove:
@moxiegirl no, we are not touching the current flow with this round of work.

markmhx:
@hstove should we close this out now that we're no longer intending to update the browser for its project?

Issue #629, opened 2017-07-18
Investigate max length of authentication responses


larrysalibra:
Some web servers say our authentication responses are too long:

<img width="1416" alt="screen shot 2017-07-19 at 3 43 01 am" src="https://user-images.githubusercontent.com/597182/28336473-b4a04b8a-6c34-11e7-8d6d-2010958b26d3.png">

PR #1875, opened 2019-03-18
Handle relative icon urls in manifest

### LINKED ISSUE #1783: Icon in app manifest are not resolved correctly
The specification for app manifests define that `icon.src` can be relative urls to the base url.
However, the browser assume they are absolute urls and then might show the wrong icon or a broken icon like this:

bildschirmfoto von 2019-01-10 09-42-57

(in this case using localhost:8888 (the browser base url) instead of the app base url)

friedger:
This PR
* uses the manifest url (if available) to resolve a relative icon url retrieved from the app manifest

* fixes #1783 

This only comes into effect after https://github.com/blockstack/blockstack.js/pull/629 is merged and the new version of blockstack.js is used - until then app developers need to specify absolute icon urls in their manifest.

Re image src (from MDN: https://developer.mozilla.org/en-US/docs/Web/Manifest#icons)
`src`     The path to the image file. If src is a relative URL, the base URL will be the URL of the manifest.

### DIFF
--- a/app/js/components/ui/containers/headers.js
+++ b/app/js/components/ui/containers/headers.js
@@ -64,6 +64,10 @@ const Header = ({
 }) => {
   const renderBugs = () => {
     if (app) {
+      const appIcon = app.manifestURI
+        ? new URL(app.icon, app.manifestURI).toString()
+        : app.icon
+
       return (
         <Bugs>
           <BlockstackBug invert={invert} size={28} />
@@ -71,7 +75,7 @@ const Header = ({
             <ChevronDoubleRightIcon color={'rgba(39, 16, 51, 0.2)'} size={18} />
           </StyledBug.Arrows>
           <Bug size={28} title={app.name}>
-            <img src={app.icon} alt={app.name} />
+            <img src={appIcon} alt={app.name} />
           </Bug>
         </Bugs>
       )


markmhx:
@friedger is this ready for review by @aulneau ?

friedger:
The PR mentioned in this issue is still open. But this is ready to review. It assumes that the manifest contains a reference to the url it was loaded from

friedger:
I added a reference in the issue description to MDN for image src in the web manifest.

timstackblock:
This still seems to serve broken image links in for some apps, the fix doesn't seem to work everywhere.

It seems to be working for me with Zinc but not with IO chat. Please see the screen shots below.

Screen Shot 2019-04-25 at 10 35 00 AM
Screen Shot 2019-04-25 at 10 34 15 AM

timstackblock:
Screen Shot 2019-04-25 at 10 48 21 AM

friedger:
Have you tested with https://github.com/blockstack/blockstack.js/pull/629?

timstackblock:
> Have you tested with [blockstack/blockstack.js#629](https://github.com/blockstack/blockstack.js/pull/629)?

Just reran the test with #629 looks good

hstove:
Because https://github.com/blockstack/blockstack.js/pull/629 is not included in the local version of blockstack.js yet, I think we'll need to wait on that before merging this.

hstove:
I'm closing this, but I merged your commit in #1922 , which includes the latest blockstack.js. Thanks!

Issue #1817, opened 2019-01-28
Ensure that "save password" modal doesn't block button on Android


markmhx:
We've learned from [previous testing](https://github.com/blockstack/blockstack-browser/issues/1775#issuecomment-452336477) that the "save password" modal on Android often blocks the button on the final step of onboarding, leading to decreased conversion.

This issue entails adjusting the button placement to ensure that it doesn't get blocked.

See previous conversation about this issue here: https://github.com/blockstack/blockstack-browser/issues/1726

![687474703a2f2f6d61726b65722e73637265656e73686f74732e70726f642e73332e616d617a6f6e6177732e636f6d2f31353439663761333032643034646662383664353138313830316639333562322d313534323734353032363932332e706e67](https://user-images.githubusercontent.com/28991/51910307-20fa7c80-23cf-11e9-8594-55e4f9bb5c0c.png)

jeffdomke:
Duplicate of an old issue. @aulneau already said there is no way to prevent it and a fix was proposed but not sure what happened to the other ticket.

markmhx:
This appears to be that other issue: https://github.com/blockstack/blockstack-browser/issues/1726

Ideally, we'd want to accommodate these kinds of password-manager popups since they're useful to the person in this context. But, of course, without blocking the continue path for people who don't want to use them.

Moving the button further up in the viewport seems like the most naive but effective solution. We have a lot of whitespace near the top that we could reduce.

markmhx:
@hstove Can you implement this one as part of the sprint starting tomorrow?

<img width="406" alt="Screen Shot 2019-03-10 at 17 40 13" src="https://user-images.githubusercontent.com/28991/54088179-9573f400-435b-11e9-91f6-e361a3d49275.png">

aulneau:
This will be resolved with https://github.com/blockstack/blockstack-browser/pull/1882

PR #1882, opened 2019-03-20
Fix onboarding button being covered on mobile screens

yknl:
Changes the UI of the last onboarding screen so that the button to continue isn't covered up on mobile screens.

image

Related issue: https://github.com/blockstack/blockstack-browser/issues/1817


jeffdomke:
@aulneau I see your commit so I assume you confirmed this works as intended when the ID displays, if not please go for it.

aulneau:
@jeffdomke Yes that's correct. It works for both states, ID and non-ID displays. This should be resolved with my commit.

PR #1889, opened 2019-03-26
Don't persist apps in redux

hstove:
- Removes the `apps` key from redux persist
- Bump redux state version

## Testing

- `git checkout v0.35.4`
- Load the app
- `git checkout feature/dont-persist-apps`
- Refresh
- Ensure state update works, redirect to homepage works


markmhx:
This was an attempted fix that we resolved another way

PR #1890, opened 2019-03-26
Fix home page apps issues

yknl:
Fixes crash on home page due to redux data.

### DIFF
--- a/app/js/HomeScreenPage.js
+++ b/app/js/HomeScreenPage.js
@@ -188,8 +188,7 @@ class HomeScreenPage extends React.Component {
   render() {
     const loading =
       this.props.apps &&
-      this.props.apps.loading &&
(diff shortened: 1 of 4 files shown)


app/js/HomeScreenPage.js
```diff
@@ -188,8 +188,7 @@ class HomeScreenPage extends React.Component {
   render() {
     const loading =
       this.props.apps &&
-      this.props.apps.loading &&
-      !this.props.apps.topApps.length
```
aulneau: We should keep this, but do this instead:

```jsx
const loading = this.props.apps && this.props.apps.loading || !this.props.apps.topApps || !this.props.apps.topApps.length
```

I believe that logic should be sound.

yknl: cool, fixed

PR #1904, opened 2019-04-09
Circleci localfix

timstackblock:
Creating a new circleci.yaml file for forked PRs


.circleci/config.yml
```diff
@@ -13,47 +13,70 @@ jobs:
...
-      TEST_E2E_GREP: login-to-hello-blockstack-app
     working_directory: ~/repo
     steps:
       - checkout
       - run: npm install
-      - run: npm run test-e2e:browserstack
+      - run:
+          name: Repo Check 1
+          command: |
+            if [ "$CIRCLE_REPOSITORY_URL" == "https://github.com/blockstack/blockstack-browser.git" ]; then
```
zone117x: Have you verified that this env var is set to the fork repo? 
Looks like there's also a way to use circleCI branch filtering for this purpose: https://discuss.circleci.com/t/create-separate-steps-jobs-for-pr-forks-versus-branches/13419/7

timstackblock: I added the Filter by branch functionality in my latest build and its a bit wonky and doesn't seem to work after doing some reading I found this article, I do not think its currently functional. https://discuss.circleci.com/t/dont-run-deploy-workflow-step-for-forked-pr/23606/4

timstackblock:
UPDATE: I was successful in splitting out each test to check if it is running on a forked repo or our repo then executing the test on Browserstack or locally. The only thing I need to do now is find a suitable docker image with all the required dependencies to run locally and we should be good to go. I am looking through the docker image repo now, if you have previously run test locally in circleCI and know a good image feel free to pass it along.

timstackblock:
Found the appropriate Docker container and the test now run locally and in Browserstack. 

But I noticed while running locally that there is a Flake that occurs quite frequently when we try to run the  login-to-hello-blockstack test locally. This is not related to the config file but the flaky test needs to be fixed before we start running locally so I am looking into that now.

.circleci/config.yml
```diff
@@ -1,59 +1,63 @@
...
-    docker:
-      - image: circleci/node:10.15.1
-    working_directory: ~/repo
-    steps:
-      - checkout
-      - run:
-          name: Install Dependencies
-          command: npm install
-      # test
-      - run: npm run test
```
zone117x: I think we still want these unit tests ran?

timstackblock: Added these back to the workflow and we have a green test run.

yknl:
@zone117x are we good to merge this one?

PR #1913, opened 2019-05-06
Update to blockstack.js v19.1.0

lgalabru:
Address https://github.com/blockstack/blockstack-browser/issues/1886


lgalabru:
@hstove thanks for looking. I haven't been experiencing this issue, but I might have a fix, and will do a pass on my personal setup.
Can you give it another try? thanks!

hstove:
Thanks, that worked and I tested out a bunch of different functions of the browser, it worked well!

PR #1920, opened 2019-05-18
update NoteRiot app icons

dantrevino:


hstove:
Hey Dan, we no longer use these images in the Browser - they're pulled directly from app.co. If you send me an image URL you want me to use, I can update it.

PR #1922, opened 2019-05-21
Handle relative icon urls in manifest

hstove:
This is the same code from @friedger in #1875 , but is built on top of the latest `develop`, which includes the latest blockstack.js stable. I made a new branch because I can't push to Friedger's fork.


friedger:
:+1:

PR #1928, opened 2019-05-30
Fix name price error.

yknl:
`/v1/prices/names` API changed, causing name price calculations to be incorrect. This causes errors during name registration due to insufficient funds. This PR switches to the `/v2/prices/names` endpoint and increases the price buffer constant from 0.0005 to 0.001 btc.

### DIFF
--- a/app/js/profiles/store/availability/actions.js
+++ b/app/js/profiles/store/availability/actions.js
@@ -6,6 +6,9 @@ import {
   isNameAvailable, getNamePrices
 } from '@utils/index'
 
+import { satoshisToBtc } from '@utils/bitcoin-utils'
+
+
 const logger = log4js.getLogger(__filename)
 
 function checkingNameAvailability(domainName) {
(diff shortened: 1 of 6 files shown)


hstove:
Thanks Ken. I've tested this locally, and it works well. Seems like there are some unit tests that need to be updated to mock the new API and use it.

app/js/profiles/store/availability/actions.js
```diff
@@ -83,7 +83,7 @@ function checkNameAvailabilityAndPrice(api, domainName) {
           dispatch(namePrice(domainName, 0))
         } else {
           return getNamePrices(api.priceUrl, domainName).then((prices) => {
-            const price = prices.total_estimated_cost.btc
+            const price = prices.name_price.satoshis * 0.000001
```
diwakergupta: @yknl what's the `0.000001` multiplier for? In general I'd recommend using symbolic names for any "magic" values to make the code more readable.

markmhx:
@kantai @jcnelson @diwakergupta Flagging this issue for you since an API change in core triggered this problem in the browser.

I'm not sure if APIs out to be versioned in core or something, but we may want to set up a process to ensure that clients (our own hosted browser or others) don't break upon core API changes without, for example, deprecation warnings.

jcnelson:
They are _definitely_ versioned.  However, `prices.total_estimated_cost.btc` was removed before we adhered to a strict versioning regime.  I remember removing this field after asking the Browser maintainers if they still used it anywhere (they said they didn't).

markmhx:
@jcnelson So we may have just had an internal comms issue here? Who were those maintainers?

Issue #1909, opened 2019-04-19
Bugs with non-default Gaia Hub & multiple accounts


zone117x:
* Bugs with multiple accounts: several areas performing gaia reads/writes use an app-state-singleton setting for the gaiahub rather than an account's specific gaiahub config. One bug causes is that only the first account can use a private gaiahub, accounts N+1 will fail to connect. 
* Default gaia hub is always used initially when restoring account, and eventually gets fixed as side-effect by later code.

timstackblock:
@jeffdomke  this is related to the .json profile issue but specifically using more that 1 Gaia hub or multiple accounts so we should take this into consideration in the rebuild adding this to your doc https://paper.dropbox.com/doc/DevX-Q3-2019--AiZQbPfKFbY1jDmfXRM8_nrhAg-QNvrtrkmUB01dI5QmQNtd

Issue #1908, opened 2019-04-19
Stale profile.json (data loss)


zone117x:
Related to https://github.com/blockstack/blockstack-browser/issues/1745
@timstackblock and I decided to open a new issue for this since the bug(s) and solution(s) may be a superset of the original issue. 

----

Several actions performed within a blockstack-browser tab cause a cached outdated profile.json to overwrite any changes made from another tab.

### Reproducible Example 
 1. Keep one or more blockstack-browser tabs are kept open (likely common user behavior).
 1. Sign into a new blockstack app that uses the publish_data scope - look up your profile.json to see the app added.
 1. Go to an open tab and update something like your name, bio, a social proof, etc. 
 1. Look up your profile.json to see the missing app entry. 

### Potential Fix
Adherence to the `single source of truth` principle. 
* Never use the cached localStorage profile object except for a few explicit cases (e.g. using the cached `name` while it is pending registration, or display UI quickly while latest data is being fetched). 
* All writes to the profile should perform a `fetch latest` -> `change only the intended profile fields` -> `upload the updated profile`.
* These write actions include:
  * Uploading avatar / profile pic.
  * Updating human name.
  * Updating bio / description.
  * Adding an app entry (during auth with publish_data). 
  * Updating a social proof or btc/eth/pgp/ssh public key. 
* The `fetch latest profile` function should always attempt a fresh resolution starting with the source of truth (the identity `address`) so that name/zonefile updates are respected. 
* All areas performing gaia read & write operations should take into consideration that more than one identity/name can be used, and each identity can have its own gaiahub configuration. Operations on the selected identity should use the gaiahub config specific to that identity. Many areas essentially use gaiahub config for `identity[0]` and/or whatever the global app state API setting happens to represent. 

#### Demo code example
```ts
async function fetchProfile(identityAddress: string) {
/**
 * First tries:
 *   1. Fetch `names` for the `address` with `bitcoinAddressLookupUrl`. 
 *   2. Fetch zonefile with `nameLookupUrl` using the first name returned. 
 *   3. Parse zonefile for the `profile.json` url. 
 * 
 * In some situations one or more parts of this resolution chain will be 
 * unavailable. Examples: 
 *   * Pending name registration (address but no name/zonefile available).
 *   * Legacy zonefile not containing a `profile.json` url. 
 *   * Legacy `profile.json` file not containing a gaiaHubUrl.
 * 
 * In these situations, the locally cached identities will be checked for a 
[...]

PR #1936, opened 2019-06-27
Identity settings in profile.json

yknl:
Adds a settings file for each identity stored on the same Gaia hub location as the profile.json. This is required for storing collection keys.


zone117x:
Is this no longer a WIP?

yknl:
No longer WIP

app/js/account/utils/index.js
```diff
@@ -116,3 +122,43 @@ export function uploadProfile(
     return uploadAttempt
   })
 }
+
+export function uploadIdentitySettings(
+  api: { gaiaHubConfig: GaiaHubConfig, gaiaHubUrl: string},
+  identityKeyPair: { key: string, keyID: string },
+  settingsData: string
```
hstove: Wondering why `settingsData` is a `string`? I would think it would be some kind of object.

app/js/utils/account-utils.js
```diff
@@ -455,6 +456,7 @@ export function getBlockchainIdentities(masterKeychain, identitiesToGenerate) {
     const identityKeyPair = deriveIdentityKeyPair(identityOwnerAddressNode)
     identityKeypairs.push(identityKeyPair)
     identityAddresses.push(identityKeyPair.address)
+    identitySettings.push({})
```
hstove: Since there is no code using these settings yet, I'm trying to reason about how this will be used. Is `identitySettings` an array, where each item is an object that represents settings for the `nth` identity in the keychain?

zone117x:
@yknl Do you know what impact this has on the stale profile.json issue https://github.com/blockstack/blockstack-browser/issues/1908, and bugs with multiple accounts + non-default gaiahub issue https://github.com/blockstack/blockstack-browser/issues/1909 ?

Code looks good, but as far as technical debt, can't tell if it's digging us deeper into those problems. 

Should we tag @timstackblock to run smoke tests on this branch?

timstackblock:
I don't think a quick smoke test would be adequate for testing issues to .json profile because we tried that before and it was broken in several places. I think a fully planned and scheduled test once we have all the pieces in place would probably be best approach. What do you guys think?

app/js/account/utils/index.js
yknl: This function expects the settings data to be already serialized. Although you could argue that the function itself should perform the serialization.

app/js/utils/account-utils.js
yknl: Yes, this is the schema: 
image

yknl:
@zone117x This PR definitely doesn't make the stale profile.json issues any better. We should address them in a separate effort.

yknl:
Agreed, will close this PR in favour of #1940

PR #1940, opened 2019-07-18
[WIP] Collections support

yknl:
This PR implements handling of collection scopes, collections storage bucket and encryption key generation based on the design proposal. It makes use of the [identity settings](https://github.com/blockstack/blockstack-browser/pull/1936) feature for storage of encryption key indexes. The key indexes are used to deterministically generate encryption keys for collection buckets. (Credits to @friedger for the suggestion) Key rotation is accomplished by incrementing the index. This reduces the impact of data corruption if randomly generated keys are used instead. This is changed from the original design where encryption keys are generated based on the list of apps granted access. 

Some open questions are whether the key generation logic should exist in blockstack.js or in the browser.


friedger:
A possible use case for collection key generation in more apps (i.e. add key generation to blockstack.js) would be to collections of data that are shared with a group of people across apps.

app/js/utils/collection-utils.js
```diff
@@ -0,0 +1,147 @@
...
+      domain: COLLECTION_GAIA_PREFIX 
+    }]
+    return connectToGaiaHub(gaiaHubUrl, collectionPrivateKey, "", scopes)
+  })
+
+  return Promise.all(hubConfigPromises)
+}
+
+function writeCollectionKeysToAppStorage(appPrivateKey, hubConfig, keyFile) {
+  const compressedAppPrivateKey = `${appPrivateKey}01`
```
hstove: Can you explain why you're doing it this way? I would think you would just use `getPublicKeyFromPrivate` from blockstack.js. This seems like you're doing some other mechanism for getting the public key.

app/js/utils/collection-utils.js
```diff
@@ -0,0 +1,145 @@
...
+const ARCHIVAL_GAIA_AUTH_SCOPE = 'putFileArchivalPrefix'
+const COLLECTION_GAIA_PREFIX = 'collection'
+
+export function getCollectionEncryptionIndex(collectionName, settings, updateIdentityCollectionSettings) {
+  if(!settings.collection 
+    || !settings.collection[collectionName] 
+    || settings.collections[collectionName].length > 0) {
+    const defaultCollectionSettings = DEFAULT_NEW_COLLECTION_SETTING_ARRAY
+    updateIdentityCollectionSettings(collectionName, defaultCollectionSettings)
+    return Promise.resolve(defaultCollectionSettings[0].encryptionKeyIndex)
```
zone117x: Trying to make sure I understand what's going on here -- is this `encryptionKeyIndex` intended to be used for bumping in order to perform a key revocation for a given collection type/scope?

yknl: Yes that's correct. This gives us a deterministic way of generating encryption keys for collections.

app/js/utils/account-utils.js
```diff
@@ -65,6 +67,59 @@ export class AppsNode {
...
+    return this.hdNode
+  }
+
+  getCollectionNode(collectionTypeName, collectionIdentifier = COLLECTION_IDENTIFIER_DEFAULT) {
+    const hash = crypto
+      .createHash('sha256')
+      .update(`${collectionTypeName}${collectionIdentifier}${this.salt}`)
+      .digest('hex')
+    const collectionIndex = hashCode(hash)
+    const collectionNode = this.hdNode.deriveHardened(collectionIndex)
```
zone117x: We should probably using a longer derivation path here. All of these inputs (`collection name`, `collection ident`, `salt`) end up as only 32 bits of uniqueness. 

This is the same problem with app-private-keys that [PR #1496](https://github.com/blockstack/blockstack-browser/pull/1496/files#diff-68143e583b120aa0075a2fdaf9f029f1R218) was trying to solve. 

This is the blockstack.js function that is able to perform better derivation, which we can leverage for collection HD nodes: https://github.com/blockstack/blockstack.js/blob/e4864ad41f376d963d4917b95996a6032f992789/src/wallet.ts#L289-L312

app/js/utils/collection-utils.js
```diff
@@ -0,0 +1,145 @@
...
+      identitySettings, 
+      updateIdentityCollectionSettings
+    )
+  const hubConfigsPromise = 
+    getCollectionGaiaHubConfigs(collectionScopes, collectionsNode, gaiaHubUrl)
+  return Promise.all([encryptionKeyPromise, hubConfigsPromise])
+    .then(results => {
+      const collectionKeys = results[0]
+      const collectionHubConfigs = results[1]
+      return updateAppCollectionKeys(
```
friedger: @yknl This adds a hard requirement to the blockstack auth flow to be online. In the current flow it is possible to sign-in while offline (for android this works offline, for the web the manifest is needed when not cached).

PR #1942, opened 2019-07-25
Upgrade blockstack.js to 19.2.1

friedger:
This PR
* upgrades blockstack.js to finally fix #1922


friedger:
Shouldn't it be 19.3.0 by now?

PR #1951, opened 2019-08-30
Update macOS native code to Swift 4.0

yknl:
Update macOS native code to Swift 4.0 so that new builds can be made with xcode 10.1+, since Apple has deprecated support for Swift 3.0 in newer xcode versions.


yknl:
Yes, the upgrade is blocking a release since the latest version of xcode deprecated support for swift 3.0.

PR #1966, opened 2019-11-07
Use blockstack.js W3C crypto

zone117x:
This PR drastically speeds up the user registration and restoration process, by using the native Web Crypto operations in `blockstack.js v21-alpha` https://github.com/blockstack/blockstack.js/issues/691.

Some sample benchmarks using a Pixel 3 XL, on both Chrome and Firefox:
* ~10 seconds when using `browser.blockstack.org`.
* ~2 seconds when using this branch.

Additional performance testing is needed with lower-end Android phones, and regression testing is needed on all platforms. 

This PR has an absurd changeset from the process of trying to debug errors while getting this working. Notably, all the synchronous code that needed to be refactored into asynchronous. 

Debugging capabilities essentially were non-existent -- like readable stack traces, breakpoints in mocha tests, breakpoints within Chrome, etc. 

Some of these notable tangental changes are:
* Completely removed Flow. Few files had Flow annotations, and its build process involvement was a major pain when debugging.
* Updated all the linting related dependencies. 
* Updated most of the toolchain dependencies (babel, webpack).

Now that the material code that actually needs changed is known, it is possible to cherrypick out only those changes. Let me know if we want to try that.


moxiegirl:
@zone117x Does this mean we have to update all the `package.json` files on the samples?

package.json
```diff
@@ -66,76 +65,69 @@
...
-    "@babel/polyfill": "^7.2.5",
-    "@babel/preset-env": "^7.3.1",
-    "@babel/preset-flow": "^7.0.0",
-    "@babel/preset-react": "^7.0.0",
-    "@babel/register": "^7.0.0",
-    "babel-eslint": "^9.0.0",
-    "babel-loader": "^8.0.5",
-    "babel-plugin-dynamic-import-node": "^2.1.0",
-    "babel-plugin-lodash": "^3.3.2",
+    "@babel/core": "^7.7.0",
```
hstove: Thank you for doing the grunt work of updating all these :+1:

test/account/store/account/actions.test.js
```diff
@@ -664,20 +664,25 @@ describe('AccountActions', () => {
       )
     })
 
-    it('dispatches buildTransaction', () => {
+    it('dispatches buildTransaction', async () => {
```
hstove: Were these changes necessary because the test wasn't actually waiting for promises?

test/require-hacks.js
```diff
@@ -10,7 +10,7 @@ requireHacker.hook('webp', () => 'module.exports = ""')
 requireHacker.hook('css', () => 'module.exports = ""')
 
 // Stub out workers
-const workers = ['encrypt', 'decrypt', 'crypto-check']
+const workers = []
```
hstove: Were these not necessary because of the webpack Workbox plugin?

app/js/utils/name-utils.js
```diff
@@ -38,7 +38,7 @@ export function isNameAvailable(lookupUrl, domainName) {
             resolve(true)
           } else {
             logger.error('isNameAvailable', response)
-            reject('Error')
+            reject(new Error(`${response.status}`))
```
hstove: Out of curiosity, what made you add this? Were you running into this somewhere?

webpack.config.js
```diff
@@ -130,25 +131,14 @@ module.exports = {
             options: { minimize: true }
           }
         ]
-      },
-      {
-        test: /\.worker\.js$/,
```
hstove: Ah, so we just aren't using Web Workers anymore, since these are now async?

hstove:
Functional QA I'm doing:

- [x] Wallet testing
  - [ ] Sending BTC
- Profile testing
  - [x] Change profile image
  - [x] Change profile info
  - [x] Github social proof (not working?)
- [x] Decrypting master keychain
- [x] Sign in with recovery code
- [x] Sign in with seed
- [x] Getting new identities
- [x] Registering a new identity on an existing keychain
- [x] Signing up from scratch


Signing in with my magic recovery code still seems kind of slow and UI blocking - though not completely. Might have just been a one-off. Signing up was quick and definitely not blocking.

Github social proofs didn't work, but they're probably already broken? I know we changed the `cheerio` import, so wanted to double check that.

hstove:
I'm seeing a weird issue with sending a BTC payment. I have $30 USD in my wallet, about 0.004 BTC, and no matter how much I try and send, I get "Not enough coin to fund fees transaction fees. Fees would be 5824, specified spend is 10". I see this comes from `blockstack.js` - is this an issue of not having the right UXTO's to put together a TX?

webpack.config.js
zone117x: Correct. Should I get rid of the rest of the workerize related dependencies and setup?

app/js/utils/name-utils.js
zone117x: The stricter linting rules indicated that non-Error types should not be thrown. Even though `throw 'some string'` is allowed in js, it tends to break dev expectations when an object in a `catch` is not an actual `Error`.

test/account/store/account/actions.test.js
zone117x: Yes, there were several instances of tests that created floating promises. The result would be a failing test with the message "did not complete", then an error stack trace in some unrelated area of the output. This was a nightmare to debug when several or more tests were failing in this way.

When these tests fail, they should now also display their associated stack trace.

.babelrc
```diff
@@ -94,9 +101,7 @@
             ],
             "alias": {
               "@components": "./app/js/components",
-              "@common": "./app/js/common",
```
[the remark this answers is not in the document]
zone117x: Correct. Breakpoints in various files would inexplicably not resolve, and stepping into functions within various files would not work and be simply skipped over. Both in the VSCode Chrome debugger and when using the regular Chrome devtools source map debugging. 

IIUC, these module import path rewrites are not working together with the rest of the tooling, like source map generation. This problem still exists for some files -- probably when a file gets bundled into one of these remaining aliased modules. 

We have these aliases specified in both `.babelrc` and `webpack.config.js`, and I assume there is some config tweaking that could get them to work, but didn't want to spend more time investigating.

zone117x:
@hstove 
> I am curious about where exactly we're getting the W3C crypto polyfill.

The W3C crypto usage comes from blockstack.js, in this PR https://github.com/blockstack/blockstack.js/pull/737

It's not really a `crypto` (Node.js module) polyfill, because there are no synchronous W3C crypto APIs available to create a polyfill. That blockstack.js PR directly uses the environment-native crypto APIs.

webpack.config.js
hstove: It wouldn't hurt to remove the dependency - is there other setup I'm missing?

zone117x:
@hstove 
> I'm seeing a weird issue with sending a BTC payment

This appears to be an existing behavior, sending small btc amounts can trigger an incorrect error message. Created an issue: https://github.com/blockstack/blockstack-browser/issues/1984

timstackblock:
@zone117x @hstove I cant send anything out of the wallet with this new release

Screen Shot 2019-12-11 at 11 27 28 AM

PR #2000, opened 2019-12-08
Add appsMeta with pubKey to profile (for storageConnected apps)

friedger:
This PR
* add `appsMeta` to the user profile (next to `apps`) that contains the public key for the user for the app.

Required for https://github.com/blockstack/blockstack.js/issues/458

### DIFF
--- a/app/js/auth/index.js
+++ b/app/js/auth/index.js
@@ -319,6 +319,11 @@ class AuthPage extends React.Component {
             apps = profile.apps
           }
 
+          let appsMeta = {}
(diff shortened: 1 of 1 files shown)


hstove:
Thanks, Friedger! I like that the benefit of this is pretty high, given not too much complexity. One thing I'll add is that we definitely need tests before merging this in. We already have tests around the `publish_data` scope, so you can probably follow those patterns to test this.

I'll bring this up in our next meeting around the user experience team.

PR #2041, opened 2020-03-11
Lossless image compression

FlipFloop:


FlipFloop:
Better called: Optimization

diwakergupta:
@FlipFloop thanks for the PR, could you help clarify exactly what problem is it intending to solve?

